PR #8, opened 2021-09-23
Architecture overview diagram

### LINKED ISSUE #9: Diagrams should be as editable as markdown code

When we want to describe a system, make a flow chart, or an activity diagram there should be a descriptive language that is used and can be edited without using another tool.

Inline detection via a GithubAction that converts plantuml into images would be great.

peternied:
Added a workflow that will automatically generate puml files into svg and commit them into the repository.

Signed-off-by: Peter Nied <petern@amazon.com>

### Issues
- Resolves #9 

### Check List
- [X] Commits are signed per the DCO using --signoff 


peternied:
See a preview with the 'rich diff' button
image

README.md
```diff
@@ -12,6 +12,14 @@
 
 OpenSearch Continous Integration is an open source CI system for OpenSearch and its plugins.
 
+### Architecture Overview
+
+![Architecture Diagram](.\img\opensearch-ci-overview.png)
```
dblock, line 17: This doesn't seem to render, need forward slashes?

README.md
```diff
@@ -12,6 +12,14 @@
...
+
+[Source](https://gist.github.com/peternied/55db1ce5277a61923519364dd6271ce5)
```
dblock, line 19: Migrate into this repo.

dblock (changes requested):

dblock:
I would think the generation job would PR the outputs when they change or otherwise maybe GHA is not the right approach here and you want some kind of precommit hook to update the SVGs?

dblock (approved):

PR #26, opened 2021-10-12
Deploy Jenkins Instance

peternied:
### Description
This is built off of the hard work from @abhinavGupta16, @gaiksaya,
and @peterzhuamazon which until now was happening behind the scenes.

This change configures and deploys Jenkins running on an EC2 instance
with recommended security practices and isolation features.  For more
details consult the readme

### Check List
- [X] Commits are signed per the DCO using --signoff 


lib/compute/jenkins-main-node.ts
```diff
@@ -0,0 +1,282 @@
+/**
```
abhinavGupta16, line 1: What happens to git history for the files that were moved? We should move the files in such a way that the history is preserved.

lib/compute/jenkins-plugins.ts
```diff
@@ -0,0 +1,155 @@
+/**
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+export class JenkinsPlugins {
```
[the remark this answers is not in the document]
peterzhuamazon: I dont see this as a problem as we want to update the plugins to the latest versions to patch all vulnerabilities.
However, I also think maintaining a list of versions for stable usage is also beneficial.
You can have 1 template for latest and 1 template for stable. What do you think about this?

lib/compute/jenkins-main-node.ts
peternied: We have a small amount of code and people working on this project, I don't think the cost is worthwhile.  The infra repository will maintain this history on its deleted files.

abhinavGupta16: Ohh, I didn't know we would have confidential information as a part of commit messages. Although , it is strange to have such info in commit messages, irrespective

peternied: In a private branch it doesn't matter what info we have in there :) so we can go wild.  We just need to be really really careful when pulling stuff out

lib/ci-stack.ts
```diff
@@ -6,21 +6,58 @@
...
+} from '@aws-cdk/core';
+import { ListenerCertificate } from '@aws-cdk/aws-elasticloadbalancingv2';
+import { CIConfigStack } from './ci-config-stack';
+import { JenkinsMainNode } from './compute/jenkins-main-node';
+import { JenkinsMonitoring } from './monitoring/ci-alarms';
+import { JenkinsExternalLoadBalancer } from './network/ci-external-load-balancer';
+import { JenkinsSecurityGroups } from './security/ci-security-groups';
 
 export class CIStack extends Stack {
   // eslint-disable-next-line no-useless-constructor
```
abhinavGupta16, line 23: This constructor is no longer useless. We can get rid of this

bin/ci-stack.ts
```diff
@@ -11,6 +9,10 @@
 import 'source-map-support/register';
 import * as cdk from '@aws-cdk/core';
 import { CIStack } from '../lib/ci-stack';
+import { CIConfigStack } from '../lib/ci-config-stack';
 
 const app = new cdk.App();
+
+new CIConfigStack(app, 'CI-Config-Dev', {});
```
abhinavGupta16, line 16: Do we need to specify the region here like we did in `Jenkins.ts`?

lib/ci-stack.ts
```diff
@@ -6,21 +6,58 @@
...
   constructor(scope: Construct, id: string, props?: StackProps) {
     super(scope, id, props);
 
-    // Network / VPC
+    const vpc = new Vpc(this, 'JenkinsVPC');
 
-    // Security Groups
+    const useSslParameter = new CfnParameter(this, 'useSsl', {
+      description: 'If the jenkins instance should be access via SSL',
+      allowedValues: ['true', 'false'],
```
abhinavGupta16, line 31: Can we use `y` or `n` instead? It would be lesser syllables to type.

abhinavGupta16 (changes requested):

lib/compute/jenkins-main-node.ts
```diff
@@ -0,0 +1,296 @@
...
+      vpcSubnets: {
+        subnetType: SubnetType.PRIVATE_WITH_NAT,
+      },
+      securityGroup: props.sg,
+      init: CloudFormationInit.fromElements(...JenkinsMainNode.configElements(
+        stack.stackName,
+        stack.region,
+        props,
+      )),
+      initOptions: { ignoreFailures: true },
```
abhinavGupta16, line 114: We probably need to remove this line.

abhinavGupta16 (changes requested):

lib/ci-stack.ts
peternied: It looks like `npm run cdk deploy CI-Dev -- --parameters useSsl=false` in practice which I think is more readable.

lib/compute/jenkins-main-node.ts
```diff
@@ -0,0 +1,296 @@
...
+      }),
+      vpc: props.vpc,
+      vpcSubnets: {
+        subnetType: SubnetType.PRIVATE_WITH_NAT,
+      },
+      securityGroup: props.sg,
+      init: CloudFormationInit.fromElements(...JenkinsMainNode.configElements(
+        stack.stackName,
+        stack.region,
+        props,
```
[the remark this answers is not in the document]
peternied: This is one of those subtle language features of typescript, super useful to make cleaner declarations

lib/ci-stack.ts
abhinavGupta16: Would it not ask as a y/n question when deploying, something like - 
`Are you sure you want to delete: JenkinsStack (y/n)? y`?

Issue #23, opened 2021-10-12
Setup Agent node config


abhinavGupta16:
### Is your feature request related to a problem? Please describe

Agent node config and keys setup is done manually currently which needs to be automated. 

### Describe the solution you'd like

Secrets Management: The setup should be automated and done using the `CI-Config-Stack` as a part of Jenkins deployment via cdk
Agent Node configuration: TBD

- [ ] ~Set up KMS key that encrypts all secrets~
- [x] Set up the secrets required by agent nodes to communicate with main node
- [x] Set up Groovy scripts that describes the agent node configuration
- [x] Set up the process to execute the groovy scripts (mostly using jenkins-cli)

PR #36, opened 2021-11-01
Add KMS key and use it for encrypting secrets

gaiksaya:

### Description
Creates a KMS key resource and uses it for encrypting secrets in AWS Secrets Manager
 
### Issues Resolved
Part of https://github.com/opensearch-project/opensearch-ci/issues/23
 
### Check List


lib/ci-config-stack.ts
```diff
@@ -24,10 +24,24 @@ export class CIConfigStack extends Stack {
...
       super(scope, id, props);
 
-      const arnSecret = new Secret(this, 'certificateArn', {});
-      const certContentsSecret = new Secret(this, 'certContents', {});
-      const privateKeySecret = new Secret(this, 'privateKey', {});
-      const redirectUrlSecret = new Secret(this, 'redirectUrl', {});
+      const kmsKey = new Key(this, 'ciKey', {
+        description: 'KMS key for encrypting ci-config secrets',
+        enableKeyRotation: true,
+        // removalPolicy: RemovalPolicy.DESTROY,
```
gaiksaya, line 30: The default retain policy is `RETAIN` for this. do we want to keep that or destroy it went the stack is destroyed?

abhinavGupta16: I think we should destroy it and then create one again when the stack is created.

abhinavGupta16:
@gaiksaya - Double checking, did the secret values on the main node EC2 instance and the certificate in load balancer fetch correctly after this change?

peternied (commented):
I think its simpler if we do not include this change.  There is always a default encryption key created by Secrets Manager in KMS when you start using it.  Let me know if there is a security concern that I am overlooking.

lib/ci-config-stack.ts
peternied: If we are making a special purpose encryption key, it should have the same lifecycle associated with it it.  The issue with retain is that it can block re-deployment since the key will already exist.

The reason that the default policy is `RETAIN` on a KMS Key is that if you were to use that key on anything else, it becomes irretrievably lost when the key vanishes, so its a guardrail - I would say `DESTORY` is appropriate for this use case.

gaiksaya:
> I think its simpler if we do not include this change. There is always a default encryption key created by Secrets Manager in KMS when you start using it. Let me know if there is a security concern that I am overlooking.

Looks it a requirement from security perspective. @peterzhuamazon is its a recommendation or requirement?

lib/ci-config-stack.ts
gaiksaya: Agreed. I'll update the PR (if if we are going ahead with this change).
Thanks!

peterzhuamazon:
> > I think its simpler if we do not include this change. There is always a default encryption key created by Secrets Manager in KMS when you start using it. Let me know if there is a security concern that I am overlooking.
> 
> Looks it a requirement from security perspective. @peterzhuamazon is it a recommendation or requirement?

Creating customer key is better than using the default because we can create separate keys for different env.

gaiksaya:
> > > I think its simpler if we do not include this change. There is always a default encryption key created by Secrets Manager in KMS when you start using it. Let me know if there is a security concern that I am overlooking.
> > 
> > 
> > Looks it a requirement from security perspective. @peterzhuamazon is it a recommendation or requirement?
> 
> Creating customer key is better than using the default because we can create separate keys for different env.

Is this something that is required or its is something that can be added as an enhancement later on?

abhinavGupta16:
> @gaiksaya - Double checking, did the secret values on the main node EC2 instance and the certificate in load balancer fetch correctly after this change?

Were you able to verify these?

gaiksaya:
> > @gaiksaya - Double checking, did the secret values on the main node EC2 instance and the certificate in load balancer fetch correctly after this change?
> 
> Were you able to verify these?

Nope did not verify. Closing this PR until we confirm if its requirement or just recommendation.

PR #37, opened 2021-11-02
changed params to read them at runtime

### LINKED ISSUE #33: [Bug]: useSsl parameter does not work
### Describe the bug

The `useSsl` parameter is always false irrespective of the input parameter value.

### To reproduce

set the `--parameter useSsl=true` and output the value of `useSsl` flag value. The value will be false.

### Expected behavior

The value of the `useSsl` boolean value should be set according to the parameter value.

### Host / Environment

_No response_

### Additional context

_No response_

### Relevant log output

_No response_

abhinavGupta16:
Signed-off-by: Abhinav Gupta <abhng@amazon.com>

### Description
The parameters were not resolved at runtime and therefore, the boolean variables were never set(always false)
 
### Issues Resolved
#33 

### Check List


lib/ci-stack.ts
```diff
@@ -33,19 +33,11 @@ export class CIStack extends Stack {
...
-
-    const runWithOidcParameter = new CfnParameter(this, 'runWithOidc', {
-      description: 'If the jenkins instance should use OIDC + federate',
-      allowedValues: ['true', 'false'],
-    });
-
-    const useSsl = useSslParameter.valueAsString === 'true';
-    const runWithOidc = runWithOidcParameter.valueAsString === 'true';
+    const runWithOidcParameter = this.node.tryGetContext('runWithOidc');
+    const runWithOidc = runWithOidcParameter === 'true';
```
gaiksaya, line 40: Glad it worked!
You need to add checks if the values entered by user is either true or false. CfnParameters does that by using `allowedValues`, for context variables we need to add checks AFAIK

abhinavGupta16: Is that provision allowed with tryGetContext()? CfnParameters could not be used since we needed the actual values at runtime.

peternied (commented):
Great work around and I love the simplicity for the command line.  When we deploy the stack, how can you track these values are set in one manner or another?

One possibility, keep the parameters but we default the value as from the cdk context so they are still recorded and would show up in a stack diff before deployment

peternied (approved):

lib/ci-stack.ts
```diff
@@ -33,19 +33,31 @@ export class CIStack extends Stack {
       },
     });
 
-    const useSslParameter = new CfnParameter(this, 'useSsl', {
-      description: 'If the jenkins instance should be access via SSL',
-      allowedValues: ['true', 'false'],
+    const useSslParameter = this.node.tryGetContext('useSsl');
+    if (useSslParameter !== 'true' && useSslParameter !== 'false') {
+      throw new ReferenceError('useSsl parameter is required to be set as "true" or "false"');
```
gaiksaya, line 38: Do we need to mention the quotes `"` as well? If not remove it. Just true or false seems fine

abhinavGupta16: I put quotes to emphasize on the 2 values that are acceptable.

gaiksaya: I think its understandable when you say it should be either true or false. The user should not get confused as to add the quotes and still see the error. Lets be clear :)

lib/ci-stack.ts
```diff
@@ -34,9 +34,17 @@ export class CIStack extends Stack {
...
     const useSslParameter = this.node.tryGetContext('useSsl');
+    if (useSslParameter !== 'true' && useSslParameter !== 'false') {
+      throw new ReferenceError('useSsl parameter is required to be set as "true" or "false"');
+    }
+
     const useSsl = useSslParameter === 'true';
 
     const runWithOidcParameter = this.node.tryGetContext('runWithOidc');
+    if (runWithOidcParameter !== 'true' && runWithOidcParameter !== 'false') {
+      throw new ReferenceError('runWithOidc parameter is required to be set as "true" or "false"');
```
peternied, line 45: Why use `ReferenceError`, just make this an `Error` type unless there is something like an argument error

abhinavGupta16: I actually used `Error` before and then later changed to `ReferenceError` since value of the variable `runWithOidcParameter` and  `useSslParameter` should only be true or false. Any other value cannot be assigned to it and I think should cause a referencing error as it should "technically" be undefined.

That being said, there isn't any rule to this that I could find. So I can put whatever everyone agrees to correct. Thoughts?

peternied: Found it https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ReferenceError

Doesn't seem correct, just use `Error` the intent isn't that someone catches it and you included the needed details so it can be remediated

peternied (approved):

lib/ci-stack.ts
```diff
@@ -33,19 +33,31 @@ export class CIStack extends Stack {
       },
     });
 
-    const useSslParameter = new CfnParameter(this, 'useSsl', {
-      description: 'If the jenkins instance should be access via SSL',
-      allowedValues: ['true', 'false'],
+    const useSslParameter = this.node.tryGetContext('useSsl');
+    if (useSslParameter !== 'true' && useSslParameter !== 'false') {
+      throw new Error('useSsl parameter is required to be set as "true" or "false"');
```
gaiksaya, line 38: Still see the quotes :) Please remove it. Errors are important and need to be clear

lib/ci-stack.ts
```diff
@@ -33,19 +33,31 @@ export class CIStack extends Stack {
...
+    }
 
-    });
+    const useSsl = useSslParameter === 'true';
+
+    const runWithOidcParameter = this.node.tryGetContext('runWithOidc');
+    if (runWithOidcParameter !== 'true' && runWithOidcParameter !== 'false') {
+      throw new Error('runWithOidc parameter is required to be set as "true" or "false"');
```
gaiksaya, line 45: Same as above

peternied (approved):

gaiksaya (approved):

PR #39, opened 2021-11-05
Add Agent Node Configuration

### LINKED ISSUE #23: Setup Agent node config

gaiksaya:

### Description
Adds agent node configuration.
The code currently adds 2 agent nodes configurations AL2-X64 and AL2-ARM64 with 100gb EBS volume. 
We can add windows, macos as an enhancement later. The number to executors (job run per instance) is 2 as of now. Let's discuss what we need to keep it as. 
 
### Issues Resolved
#23 
 
### Check List


lib/ci-stack.ts
```diff
@@ -80,6 +90,33 @@ export class CIStack extends Stack {
       oidcCredArn: importedOidcConfigValuesSecretBucketValue.toString(),
       useSsl,
       runWithOidc,
+    },
+    {
+      region: this.region.toString(),
+      agent_node_security_group: securityGroups.agentNodeSG.securityGroupId.toString(),
+      subnet_id: vpc.publicSubnets[0].subnetId.toString(),
+    },
+    {
```
peternied, line 99: Lets move these into the `AgentNodeConfig` file, otherwise this `Stack` contains a whole lot of little detail that makes it hard to understand the pieces that are going into the stack.  It could look something like:

```
export class AgentNodeConfig {
   static readonly AGENT_CONFIGS: CloudAgentConfig[] = [{
      ec2_cloud_name: 'AL2-X64',
      ...
   }]
}
```

This way when we add update the configuration only one clearly organized place is updated.

lib/ci-stack.ts
```diff
@@ -80,6 +90,33 @@ export class CIStack extends Stack {
       oidcCredArn: importedOidcConfigValuesSecretBucketValue.toString(),
       useSsl,
       runWithOidc,
+    },
+    {
+      region: this.region.toString(),
```
peternied, line 95: region does not need to be passed in, it can be extracted from `stack.region`

lib/compute/jenkins-main-node.ts
```diff
@@ -329,6 +349,19 @@ export class JenkinsMainNode {
       // Therefore, sleep 60 seconds to wait for plugins to install and jenkins to start which is required for the next step
       InitCommand.shellCommand('sleep 60'),
 
+      // Create groovy script that holds the agent Node config for EC2 plugin ref:https://gist.github.com/vrivellino/97954495938e38421ba4504049fd44ea
+      agentNode.asInitFile('/agent-node-al2-x64.groovy', agentNodeProps, al2x64AgentNodeConfig),
```
peternied, line 353: Can we pass in an `AgentNode` object with a method that produces these InitElement and adds them to the end of the list?  It would reduce the amount of argument passing by a wide margin and reduce coupling

line 137:
```
CloudFormationInit.fromElements(...JenkinsMainNode.configElements(...),
                                ...props.cloudAgent.configElements())
```

lib/security/ci-security-groups.ts
```diff
@@ -26,7 +26,13 @@ export class JenkinsSecurityGroups {
       vpc,
       description: 'Main node of Jenkins',
     });
+
+    this.agentNodeSG = new SecurityGroup(stack, 'AgentNodeSG', {
```
peternied, line 30: move the declaration and the additional rules after the main agent nodes, so it reads more like 

* External Access SG / Rules 
* MainNode SG / Rules
* AgentNode SG / Rules

lib/compute/jenkins-main-node.ts
```diff
@@ -34,6 +35,14 @@ export interface JenkinsMainNodeProps extends HttpConfigProps, OidcFederateProps
   readonly sg: SecurityGroup;
 }
 
+export interface AgentNodeProps{
+  readonly region : string;
+  // eslint-disable-next-line camelcase
```
peternied, line 40: This interface should follow our normal rules since it isn't `toJSON`ed

lib/compute/agent-nodes.ts
```diff
@@ -0,0 +1,185 @@
...
+      assumedBy: new ServicePrincipal('ec2.amazonaws.com'),
+    });
+    const AgentNodeInstanceProfile = new CfnInstanceProfile(stack, 'JenkinsAgentNodeInstanceProfile', { roles: [AgentNodeRole.roleName] });
+    this.AgentNodeInstanceProfileArn = AgentNodeInstanceProfile.attrArn.toString();
+    this.SSHEC2KeySecretId = Fn.join('/', ['ec2-ssh-key', key.keyPairName.toString(), 'private']);
+  }
+
+  public asInitFile(filePath: string, stackConfig: AgentNodeProps, config: AgentNodeConfig, options?: InitFileOptions): InitFile {
+    return InitFile.fromString(filePath,
+      `import hudson.model.*
```
peternied, line 52: I know this is a big hunk of groovy in a string; it would be easier to read if we carefully indented and aligned columns just like it was our typescript code.

lib/compute/agent-nodes.ts
```diff
@@ -0,0 +1,185 @@
...
+    import jenkins.model.*
+    import hudson.plugins.ec2.*
+    import com.amazonaws.services.ec2.model.InstanceType
+
+    def instance = Jenkins.getInstance()
+    def init_script = 'sudo mkdir -p /var/jenkins/ && sudo chown -R ec2-user:ec2-user /var/jenkins \
+    && sudo yum install -y java-1.8.0-openjdk cmake python3 python3-pip && sudo yum groupinstall -y "Development Tools" \
```
peternied, line 59: Since this are line continuations please indent, occurs in a number of places

lib/compute/agent-nodes.ts
```diff
@@ -0,0 +1,185 @@
...
+    && sudo ln -sfn \`which pip3\` /usr/bin/pip && pip3 install pipenv && sudo ln -s ~/.local/bin/pipenv /usr/local/bin'
+
+// variables
```
peternied, line 62: Align the comment with the definition, and it isn't very clear to me based on naming the difference between this grouping and the one below it.

lib/compute/agent-nodes.ts
```diff
@@ -0,0 +1,185 @@
...
+
+// info imported the stack
+    def region = "${stackConfig.region}"
+    def security_group = "${stackConfig.agent_node_security_group}"
+    def subnet_id = "${stackConfig.subnet_id}"
+    def instance_profile_arn = "${this.AgentNodeInstanceProfileArn}"
+    def sshKeysASMSecretId = "${this.SSHEC2KeySecretId}"
+    def ec2_tags = [new EC2Tag('Name', 'jenkins-agent-node')]
+
+    def agent_node_template = new SlaveTemplate(
```
peternied, line 78: This code is tightly coupled to `SlaveTemplate` could you link where this interface is described so future maintainers can use it for follow up?

lib/compute/agent-nodes.ts
```diff
@@ -0,0 +1,185 @@
...
+      region,
+      // String privateKey
+      null,
+      // String sshKeysCredentialsId
+      sshKeysASMSecretId,
+      // String instanceCapStr
+      "10",
+      // List<? extends SlaveTemplate> templates
+      [agent_node_template],
+      // String roleArn
```
peternied, line 176: Is this the role of the running agent node? I think we should have one so we can use SSM to ssh into the host just like the [main node](https://github.com/opensearch-project/opensearch-ci/blob/86aea2e78a29cd02e1df77031a1b0b57e09f2dd9/lib/compute/jenkins-main-node.ts#L70) for troubleshooting.

lib/compute/agent-nodes.ts
peternied: Note; also applies to the AmazonEC2Cloud constructor below.

lib/compute/agent-nodes.ts
```diff
@@ -0,0 +1,185 @@
...
+ */
+
+import { CfnInstanceProfile, Role, ServicePrincipal } from '@aws-cdk/aws-iam';
+import { Fn, Stack, Tags } from '@aws-cdk/core';
+import { KeyPair } from 'cdk-ec2-key-pair';
+import { InitFile, InitFileOptions } from '@aws-cdk/aws-ec2';
+import { AgentNodeProps } from './jenkins-main-node';
+
+export interface AgentNodeConfig{
+  // eslint-disable-next-line camelcase
```
peternied, line 16: This interfaces style should follow normal conventions

lib/compute/agent-nodes.ts
peternied: - it looks like the instance profile has a role, maybe a comment would be useful instead

lib/compute/agent-nodes.ts
peternied, line 59: Please break apart into one operation per line for ease of reading

peternied: Actually, could you declare this out side of the groovy and import it?

lib/ci-stack.ts
gaiksaya: The method is not a part of the constructor so need to pass that as a parameter https://github.com/opensearch-project/opensearch-ci/pull/39/files#diff-90fd20658ffdeb68fbac6a146498601102a44c741138481202ff494189e69c30R50
Should I pass stack as a param then and then extract region from it?

peternied: Sure, but do that at the last possible minute in the code so you aren't passing that parameter around when it is unused.

lib/compute/agent-nodes.ts
abhinavGupta16: Alternatively, would it be better to maybe write the groovy code as a groovy file and import it? 
Values can then later be replaced as variables from the props. This would make the groovy code more readable and IDEs would also be able to parse it to verify for errors etc

gaiksaya:
> @abhinavGupta16 Alternatively, would it be better to maybe write the groovy code as a groovy file and import it?
Values can then later be replaced as variables from the props. This would make the groovy code more readable and IDEs would also be able to parse it to verify for errors etc<
I did think about that, about using [InitFile.fromAsset](https://docs.aws.amazon.com/cdk/api/latest/docs/@aws-cdk_aws-ec2.InitFile.html#static-fromwbrassettargetfilename-path-options) but how would we change the values once the file is created in the ec2? These files contain dynamic values like SG, subnet ids, etc. How should we replace that?

peternied:
> > @abhinavGupta16 Alternatively, would it be better to maybe write the groovy code as a groovy file and import it?
> > Values can then later be replaced as variables from the props. This would make the groovy code more readable and IDEs would also be able to parse it to verify for errors etc
> 
> I did think about that, about using [InitFile.fromAsset](https://docs.aws.amazon.com/cdk/api/latest/docs/@aws-cdk_aws-ec2.InitFile.html#static-fromwbrassettargetfilename-path-options) but how would we change the values once the file is created in the ec2 or even before it? These files contain dynamic values like SG, subnet ids, etc. How should we replace that? Will look into it if we can use `Cfn:Fn:sub` like I used it [here](https://github.com/opensearch-project/opensearch-infra/blob/main/opensearch-cluster/cdk/multi-node/multi_node_cdk/infra.py#L167)

I would recommend not spending too many cycles here, while not ideal there is minimal functionality in that groovy code that is cannot be validated other than seeing a fresh main node spin up.  If we have more ideas lets consider making issues for them to resolve over time.

abhinavGupta16:
> > > @abhinavGupta16 Alternatively, would it be better to maybe write the groovy code as a groovy file and import it?
> > > Values can then later be replaced as variables from the props. This would make the groovy code more readable and IDEs would also be able to parse it to verify for errors etc
> > 
> > 
> > I did think about that, about using [InitFile.fromAsset](https://docs.aws.amazon.com/cdk/api/latest/docs/@aws-cdk_aws-ec2.InitFile.html#static-fromwbrassettargetfilename-path-options) but how would we change the values once the file is created in the ec2 or even before it? These files contain dynamic values like SG, subnet ids, etc. How should we replace that? Will look into it if we can use `Cfn:Fn:sub` like I used it [here](https://github.com/opensearch-project/opensearch-infra/blob/main/opensearch-cluster/cdk/multi-node/multi_node_cdk/infra.py#L167)
> 
> I would recommend not spending too many cycles here, while not ideal there is minimal functionality in that groovy code that is can be validated other than starting up a fresh main node and validating it. If we have more ideas lets consider making issues for them to resolve seperately.

@peternied - We can definitely create an issue and do it the separate cycle. 
@gaiksaya - A few says I can think of is taking variables as a parameter to the script and running it. Alternatively, we can read the entire file as a string in the typescript code and replace the variable values. Another, not so nice method could be saving the file on EC2 and replace the values on the file using cmd.

lib/ci-stack.ts
```diff
@@ -6,7 +6,17 @@
  * compatible open source license.
  */
 
-import { FlowLogDestination, FlowLogTrafficType, Vpc } from '@aws-cdk/aws-ec2';
+import * as ec2 from '@aws-cdk/aws-ec2';
```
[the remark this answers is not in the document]
gaiksaya: Corrected

lib/ci-stack.ts
gaiksaya: Done

lib/ci-stack.ts
gaiksaya: Done

lib/security/ci-security-groups.ts
gaiksaya: Done

gaiksaya: Done

lib/compute/agent-nodes.ts
gaiksaya: Added

lib/compute/agent-nodes.ts
gaiksaya: Added

lib/compute/agent-nodes.ts
gaiksaya: Done

lib/compute/agent-nodes.ts
```diff
@@ -0,0 +1,185 @@
...
+  ami_id : string;
```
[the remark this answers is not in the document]
gaiksaya: Done

package-lock.json
```diff
@@ -1,13190 +1,8 @@
 {
   "name": "opensearch-ci-stack",
   "version": "0.1.0",
-  "lockfileVersion": 2,
+  "lockfileVersion": 1,
```
peternied, line 4: Can you revert this change?

peternied (commented):
Looking really good, should it move out of draft or did you have more changes in mind?

gaiksaya:
> Looking really good, should it move out of draft or did you have more changes in mind?

All tests are broken. And I missed one of your comments which will really clean things up

>Can we pass in an AgentNode object with a method that produces these InitElement and adds them to the end of the list? It would reduce the amount of argument passing by a wide margin and reduce coupling
line 137:
CloudFormationInit.fromElements(...JenkinsMainNode.configElements(...),
                                ...props.cloudAgent.configElements())


So still working on it.

peternied:
> > > > @abhinavGupta16 Alternatively, would it be better to maybe write the groovy code as a groovy file and import it?
> > > > Values can then later be replaced as variables from the props. This would make the groovy code more readable and IDEs would also be able to parse it to verify for errors etc
> > > 
> > > 
> > > I did think about that, about using [InitFile.fromAsset](https://docs.aws.amazon.com/cdk/api/latest/docs/@aws-cdk_aws-ec2.InitFile.html#static-fromwbrassettargetfilename-path-options) but how would we change the values once the file is created in the ec2 or even before it? These files contain dynamic values like SG, subnet ids, etc. How should we replace that? Will look into it if we can use `Cfn:Fn:sub` like I used it [here](https://github.com/opensearch-project/opensearch-infra/blob/main/opensearch-cluster/cdk/multi-node/multi_node_cdk/infra.py#L167)
> > 
> > 
> > I would recommend not spending too many cycles here, while not ideal there is minimal functionality in that groovy code that is can be validated other than starting up a fresh main node and validating it. If we have more ideas lets consider making issues for them to resolve seperately.
> 
> @peternied - We can definitely create an issue and do it the separate cycle. The reason I ask is because this would make our code more readable and maintainable.

Making the largest thread I can! For your consideration, readability and maintainability are features that have a time cost, sometimes we just need action (consider the effort and follow up tasks around docker generation) other times we need careful construction for long term payoff (consider our internal deployment system, its hardly even started and we are still planning it!) it is always important to ask how likely the time will return dividends.

Supporting groovy code in this project, as is, doesn't seem very fruitful to me; however, if we had a use case like DB is addressing in https://github.com/opensearch-project/opensearch-build/pull/924 a systemic problem that has hampered us many times.  We should always make sure we have a goal in mind and are aware of the constraints we place on what we build.

lib/compute/jenkins-main-node.ts
gaiksaya: Done

gaiksaya:
Opened up issue for loading groovy scripts from files: https://github.com/opensearch-project/opensearch-ci/issues/44

I have not added tests for agent-nodes specifically as methods are not static and need to instatiate the class which is causing me to run into 
```
 There is already a Construct with name 'AgentNode-KeyPair' in CIStack [MyTestStack]
```

Not sure if tests are required for agent-nodes and they itself are dynamic.

package-lock.json
gaiksaya: This should be 1 or 2?

README.md
```diff
@@ -44,6 +44,7 @@ OpenSearch Continuous Integration is an open source CI system for OpenSearch and
 6. Log onto the AWS Console of the account, navigate to [cloud watch](https://console.aws.amazon.com/cloudwatch/home), open log groups, looking for `JenkinsMainNode/var/log/jenkins/jenkins.log`
 7. Search the logs for `Jenkins initial setup is required. An admin user has been created and a password generated.` After that entry the password for the jenkins instance will be in the cloudwatch logs.
 8. Go to the `CI-Dev.JenkinsExternalLoadBalancerDns` url returned by CDK output to access the jenkins host.
+9. While destroying the stack make sure you delete the agent nodes manually (via jenkins UI or AWS console) so that shared resources (like vpc, security groups, etc) can be deleted.
```
peternied, line 47: Switch to -> `If you destroy the stack,` I never destroy my stacks, and I wouldn't recommend it to most developers unless you are doing specific setup where you need a fresh environment.

lib/ci-stack.ts
```diff
@@ -80,7 +85,13 @@ export class CIStack extends Stack {
       oidcCredArn: importedOidcConfigValuesSecretBucketValue.toString(),
       useSsl,
       runWithOidc,
-    });
+    },
+    {
+      agentNodeSecurityGroup: securityGroups.agentNodeSG.securityGroupId.toString(),
```
peternied, line 90: Don't toString these, you can pass around the security group and have it tostring before its used if needed

lib/ci-stack.ts
```diff
@@ -80,7 +85,13 @@ export class CIStack extends Stack {
...
+      subnetId: vpc.publicSubnets[0].subnetId.toString(),
+    },
+    agentNodesConfig.AL2_X64,
```
peternied, line 93: Can you have the main node accept an agentNodesConfig and then it extracts what it needs?

lib/ci-stack.ts
peternied, line 90: Seems like the agent configuration should be passed into the CloudAgentNodeConfig object instead of into the jenkins main node.

lib/compute/agent-node-config.ts
```diff
@@ -0,0 +1,44 @@
...
+
+  readonly AL2_ARM64: AgentNodeConfig;
+
+  constructor(stack: Stack) {
+    this.AL2_X64 = {
+      ec2CloudName: 'AL2-X64',
+      instanceType: InstanceType.of(InstanceClass.C5, InstanceSize.XLARGE4).toString(),
+      workerLabelString: 'AL2-X64',
+      numberOfExecutors: '2',
+      remoteUser: 'ec2-user',
```
peternied, line 26: Any reason we can't hardcode this value?

lib/compute/agent-node-config.ts
```diff
@@ -0,0 +1,44 @@
...
+      amiId: MachineImage.latestAmazonLinux({
+        generation: AmazonLinuxGeneration.AMAZON_LINUX_2,
+        cpuType: AmazonLinuxCpuType.X86_64,
+      }).getImage(stack).imageId.toString(),
+    };
+    this.AL2_ARM64 = {
+      ec2CloudName: 'AL2-ARM64',
+      instanceType: InstanceType.of(InstanceClass.C6G, InstanceSize.XLARGE4).toString(),
+      workerLabelString: 'AL2-ARM64',
+      numberOfExecutors: '2',
```
peternied, line 36: Can we hardcode this value too?

lib/compute/agent-nodes.ts
```diff
@@ -0,0 +1,210 @@
...
+      '',
+      // String roleSessionName
+      ''
+    )
+
+    instance.clouds.add(new_cloud)
+    instance.save()`);
+  }
+
+  public configElements(stackRegion: string, agentNodeProps: AgentNodeProps, al2x64AgentNodeConfig: AgentNodeConfig,
```
peternied, line 192: These props seem like they can be fields of this class rather than being passed around

lib/compute/jenkins-main-node.ts
```diff
@@ -34,6 +35,11 @@ export interface JenkinsMainNodeProps extends HttpConfigProps, OidcFederateProps
   readonly sg: SecurityGroup;
 }
 
+export interface AgentNodeProps{
```
peternied, line 38: Move to the agent node class

peternied (approved):
Feedback is mostly nitpicking to improve encapsulation, looks pretty good.

lib/compute/agent-node-config.ts
gaiksaya: I am using same groovy script to install 2 types of agent nodes inorder to avoid duplication. Maybe we can refactor later when we implement #44 . So we can have multiple groovy scripts for each type of agent node

lib/compute/agent-node-config.ts
gaiksaya: Same as above

lib/ci-stack.ts
gaiksaya: Done

lib/ci-stack.ts
gaiksaya: Done

lib/compute/jenkins-main-node.ts
gaiksaya: I did, Forgot to remove from here. Removed now

lib/compute/agent-nodes.ts
gaiksaya: Will refactor later is required. Getting this PR in for now.

gaiksaya:
@peternied Are you okay with me merging this PR? Any objection?

abhinavGupta16 (approved):

peternied:
@gaiksaya I've already signed off you can merge whenever you'd like.

If you were open to changes, I'd recommend reverting the changes to `package-lock.json`

gaiksaya:
> @gaiksaya I've already signed off you can merge whenever you'd like.
> 
> If you were open to changes, I'd recommend reverting the changes to `package-lock.json`

Is it just `lockfileVersion`? or entire package-lock.json?

abhinavGupta16:
> > @gaiksaya I've already signed off you can merge whenever you'd like.
> > If you were open to changes, I'd recommend reverting the changes to `package-lock.json`
> 
> Is it just `lockfileVersion`? or entire package-lock.json?

I don't think you revert the entire file since you have a change in `package.json` which should now reflect in `package-lock.json`

gaiksaya:
> > @gaiksaya I've already signed off you can merge whenever you'd like.
> > If you were open to changes, I'd recommend reverting the changes to `package-lock.json`
> 
> Is it just `lockfileVersion`? or entire package-lock.json?

Yeah since the file is auto generated concerned if I should change anything at all

gaiksaya:
> > > @gaiksaya I've already signed off you can merge whenever you'd like.
> > > If you were open to changes, I'd recommend reverting the changes to `package-lock.json`
> > 
> > 
> > Is it just `lockfileVersion`? or entire package-lock.json?
> 
> Yeah since the file is auto generated , I am concerned if I should change anything at all

Update: Reverted the package-lock.json changes and ran `npm run build` the project is still building. Merging this PR.
Thanks!

PR #63, opened 2022-02-11
Add secrets description and fix README

gaiksaya:

### Description
Add secrets description so that it is easy to update them via console. Also fixes a missed step in README due to which jenkins with SSL feature was failing.
 
### Issues Resolved
 
### Check List


lib/ci-config-stack.ts
```diff
@@ -28,12 +28,24 @@ export class CIConfigStack extends Stack {
...
-      const certContentsSecret = new Secret(this, 'certContents', {});
-      const certChainSecret = new Secret(this, 'certChain', {});
-      const privateKeySecret = new Secret(this, 'privateKey', {});
-      const redirectUrlSecret = new Secret(this, 'redirectUrl', {});
-      const OIDCConfigValuesSecret = new Secret(this, 'OIDCConfigValues', {});
+      const arnSecret = new Secret(this, 'certificateArn', {
+        description: 'Certificate ARN retrieved after uploading certificate to IAM server',
+      });
+      const certContentsSecret = new Secret(this, 'certContents', {
+        description: 'Contents of SSL certificate',
```
peterzhuamazon, line 35: Suggest: Public Key

gaiksaya: Fixed. Thanks for the suggestion.

peterzhuamazon (approved):

lib/ci-config-stack.ts
```diff
@@ -28,12 +28,24 @@ export class CIConfigStack extends Stack {
...
+        description: 'Contents of public key of the SSL certificate',
+      });
+      const certChainSecret = new Secret(this, 'certChain', {
+        description: 'Contents of the SSL certificate chain',
+      });
+      const privateKeySecret = new Secret(this, 'privateKey', {
+        description: 'Contents of private key of the SSL certificate',
+      });
+      const redirectUrlSecret = new Secret(this, 'redirectUrl', {
+        description: 'Redirect url for Jenkins',
```
abhinavGupta16, line 44: I think it's a good idea to add a default value as well for this secret as we discussed offline.

abhinavGupta16 (approved):

lib/ci-config-stack.ts
gaiksaya: I did not find a direct way of doing that. I see this feature https://docs.aws.amazon.com/cdk/api/v1/docs/@aws-cdk_aws-secretsmanager.SecretStringValueBeta1.html which was implemented `v1.138.0` and later

PR #70, opened 2022-02-16
Change the minimum number of instances

gaiksaya:
### Description
Change the minimum number of instances to zero which is infinite.
 
### Issues Resolved
 
### Check List


peterzhuamazon (approved):
Thanks.

abhinavGupta16 (approved):

abhinavGupta16:
@gaiksaya - Work for future, we would eventually want to pull these properties from a properties files which can be environment specific. It would give us a lot more control and flexibility. thoughts?

gaiksaya:
> @gaiksaya - Work for future, we would eventually want to pull these properties from a properties files which can be environment specific. It would give us a lot more control and flexibility. thoughts?

We do have this values as properties here: https://github.com/opensearch-project/opensearch-ci/blob/main/lib/compute/agent-node-config.ts#L20-L41
However the common config is hard coded in the template. If we feel any value which should differ based on the agent node type, we should make it as property like others.

Issue #52, opened 2022-01-07
CDK deployment creates role and policy for ECR support


abhinavGupta16:
### Is your feature request related to a problem? Please describe

Add support to create a role which can log into `ecr` and push docker images.

### Describe the solution you'd like

Create cdk support to be able to create the below policy and add trust relationship with the agent node and main node.

```
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Allow",
            "Action": [
                "ecr-public:BatchCheckLayerAvailability",
                "ecr-public:CompleteLayerUpload",
                "ecr-public:InitiateLayerUpload",
                "ecr-public:PutImage",
                "ecr-public:UploadLayerPart"
            ],
            "Resource": "*"
        },
        {
            "Effect": "Allow",
            "Action": [
                "ecr-public:GetAuthorizationToken",
                "sts:GetServiceBearerToken"
            ],
            "Resource": "*"
        }
    ]
}
```

PR #77, opened 2022-02-24
Assets stack to deploy ECR

abhinavGupta16:
### Description
CDK is being used to deploy roles, policies and public ecr repositories
 
### Issues Resolved
https://github.com/opensearch-project/opensearch-ci/issues/52
 
### Check List


abhinavGupta16:
Currently facing issues with docker on jenkins deployed with opensearch-ci. I am therefore yet to test out these changes by running a job

bin/ci-stack.ts
```diff
@@ -7,12 +7,28 @@
...
-import { App } from '@aws-cdk/core';
+import { App, RemovalPolicy } from '@aws-cdk/core';
 import { CIStack } from '../lib/ci-stack';
 import { CIConfigStack } from '../lib/ci-config-stack';
+import { DeployAWSAssets } from '../lib/DeployAWSAssets';
 
 const app = new App();
 
-new CIConfigStack(app, 'CI-Config-Dev', {});
+const env: string = 'Dev';
```
gaiksaya, line 17: Why are we hard coding the env?

bin/ci-stack.ts
```diff
@@ -7,12 +7,28 @@
...
 
-new CIStack(app, 'CI-Dev', { });
+const ciConfigStack = new CIConfigStack(app, `OpenSearch-CI-Config-${env}`, {});
+
+const ciStack = new CIStack(app, `OpenSearch-CI-${env}`, {
+  ecrAccountId: ciConfigStack.account,
+});
+
+new DeployAWSAssets(app, `OpenSearch-CI-Deploy-Assets-${env}`, {
+  removalPolicy: RemovalPolicy.DESTROY,
```
gaiksaya, line 26: We do not want the Assests to be destroyed not matter what. Please remove this policy assuming default is retain

bin/ci-stack.ts
```diff
@@ -7,12 +7,28 @@
...
+  mainNodeAccountNumber: ciStack.account,
+  createEcrRepositories: true,
+  envName: env,
+  env: {
+    // public ECR repositories can only be created in us-east-1
```
gaiksaya, line 31: Can you also comment the issue that you found stating this fact so that we can track in future?

lib/ci-ecr-stack.ts
```diff
@@ -0,0 +1,87 @@
...
+      CiEcrStack.createRepositories(scope, this, props.removalPolicy ?? RemovalPolicy.RETAIN);
+    }
+    const ecrPolicy = CiEcrStack.createEcrPolicy(this, id);
+
+    CiEcrStack.createEcrRole(this, ecrPolicy, props.mainNodeAccountNumber);
+  }
+
+  public static createEcrRole(stack: Stack, ecrPolicy: ManagedPolicy, mainNodeAccountNumber: string) : IRole {
+    return new Role(stack, 'ecr-stack-role', {
+      roleName: 'OpenSearch-CI-ECR-ecr-role',
```
gaiksaya, line 33: Once ECR is enough to say its an ECR role, rename to `OpenSearch-CI-ecr-role`

lib/ci-stack.ts
```diff
@@ -31,12 +31,14 @@ export interface CIStackProps extends StackProps {
   runWithOidc?: boolean;
   /** Additional verification during deployment and resource startup. */
   strictMode?: boolean;
+  /** Environment the stack is ebing deployed on */
```
gaiksaya, line 34: Typo: being. Also I think this should be reframed as `Account to deploy your ECR Assets on`

lib/compute/jenkins-main-node.ts
```diff
@@ -76,6 +82,44 @@ export class JenkinsMainNode {
       }),
     };
 
+    const agentNode = new AgentNode(stack);
+    this.ec2Instance = new Instance(stack, 'MainNode', {
+
+      instanceType: InstanceType.of(InstanceClass.C5, InstanceSize.XLARGE4),
```
gaiksaya, line 88: Why the restructuring?

gaiksaya (changes requested):
Please fix the failing tests

bin/ci-stack.ts
abhinavGupta16: This is hardcoded only for `opensearch-ci` here. This file does not exist in the other deployment

bin/ci-stack.ts
abhinavGupta16: The default policy is Retain. I mentioned destroy here since this deployment is only used for dev envs. https://github.com/opensearch-project/opensearch-ci/pull/77/files#diff-d638b1fcda588e6ca1f48230caf1d17595dccafad1e580bcd937a15150343471R24

lib/ci-ecr-stack.ts
abhinavGupta16: `OpenSearch-CI-ECR` is the stack name and ecr-role is the role name

lib/ci-stack.ts
abhinavGupta16: It is not the account, but the env. Account number is a different property. I'll fix the typo

lib/compute/jenkins-main-node.ts
abhinavGupta16: There were a lot of policies that were being created in the constructor and then each of them added individually in a new line. This implementation is much neater and easier to extend in future if we add a new policy. We can now simply go and create a policy in the create policy function and it will be added to the role on ec2

bin/ci-stack.ts
gaiksaya: Not necessarily. We say that this code is production if any external users want to deploy the stack. If they are using this code base it will easily be removed. Better retain like we have for other assets like s3 auditing bucket, etc.

lib/ci-ecr-stack.ts
gaiksaya: Why not grab the stack name from `stack.name` property then?

bin/ci-stack.ts
gaiksaya: Let think about this project as an individual entity than linking to another deployment.

abhinavGupta16: Considering this as an individual entity, this repository would be used for dev and therefore, dev

bin/ci-stack.ts
abhinavGupta16: This code is production, but by no means is these are the properties used in production. If we retain the assets, the redeployment fails since cloud formation does not update the ecr repository resource and would require user to delete everything manually each time.

lib/ci-ecr-stack.ts
abhinavGupta16: `stack.name` would also contain the env where the stack is being deployed. Which would cause the role name to change according to the env, creating a problem when using roles in `jenkinsfile`

abhinavGupta16:
> Please fix the failing tests

Fixed and added new test cases as I marked this ready for review

bin/ci-stack.ts
gaiksaya: This repository is not dev let's be clear on that. :) We should assume each repository/code is production grade and contribute accordingly.

bin/ci-stack.ts
gaiksaya: Thats fine, the user can manually delete it but here is a use case that if we delete the stack by mistake or something happens, we want to be retain any ECR repo since those are public. That should never be taken down since it's one of the release channel now.

gaiksaya (changes requested):

abhinavGupta16:
> Thats fine, the user can manually delete it but here is a use case that if we delete the stack by mistake or something 
> happens, we want to be retain any ECR repo since those are public. That should never be taken down since it's one of the
> release channel now.

The default is `Retain`

abhinavGupta16:
> This repository is not dev let's be clear on that. :) We should assume each repository/code is production grade and contribute accordingly.

What should be the default value for this?

gaiksaya:
> > Thats fine, the user can manually delete it but here is a use case that if we delete the stack by mistake or something
> > happens, we want to be retain any ECR repo since those are public. That should never be taken down since it's one of the
> > release channel now.
> 
> The default is `Retain`. The public repos will be created with Retain. This overrides the default.

We do not want to override it. Keep the default.

abhinavGupta16:
> > > Thats fine, the user can manually delete it but here is a use case that if we delete the stack by mistake or something
> > > happens, we want to be retain any ECR repo since those are public. That should never be taken down since it's one of the
> > > release channel now.
> > 
> > 
> > The default is `Retain`. The public repos will be created with Retain. This overrides the default.
> 
> We do not want to override it. Keep the default.

I still think we shouldn't. Let's agree to disagree

gaiksaya:
> > > > Thats fine, the user can manually delete it but here is a use case that if we delete the stack by mistake or something
> > > > happens, we want to be retain any ECR repo since those are public. That should never be taken down since it's one of the
> > > > release channel now.
> > > 
> > > 
> > > The default is `Retain`. The public repos will be created with Retain. This overrides the default.
> > 
> > 
> > We do not want to override it. Keep the default.
> 
> I still think we shouldn't. It's a lot of work to delete the repository every time with every redeployment for the stack. Again, this is not what is being used, or will be used for production. Let's agree to disagree

I have strong opinion about this since assets are very crucial and are release platform we do not want accidental deletions to affect the assets. Right now we even have S3 bucket that is retained after each deployment but that is fine. https://github.com/opensearch-project/opensearch-ci/blob/main/lib/auditing/ci-audit-logging.ts#L16
We can get more input from @peterzhuamazon @zelinh @bbarani

gaiksaya:
> > > > > Thats fine, the user can manually delete it but here is a use case that if we delete the stack by mistake or something
> > > > > happens, we want to be retain any ECR repo since those are public. That should never be taken down since it's one of the
> > > > > release channel now.
> > > > 
> > > > 
> > > > The default is `Retain`. The public repos will be created with Retain. This overrides the default.
> > > 
> > > 
> > > We do not want to override it. Keep the default.
> > 
> > 
> > I still think we shouldn't. It's a lot of work to delete the repository every time with every redeployment for the stack. Again, this is not what is being used, or will be used for production. Let's agree to disagree
> 
> I have strong opinion about this since assets are very crucial and are release platform we do not want accidental deletions to affect the assets. Right now we even have S3 bucket that is retained after each deployment but that is fine as well since we want the audit to be there even if the stack is deployed. https://github.com/opensearch-project/opensearch-ci/blob/main/lib/auditing/ci-audit-logging.ts#L16 We can get more input from @peterzhuamazon @zelinh @bbarani

Update: @abhinavGupta16 is updating readme file stating the consequences of dev stack.

lib/ci-ecr-stack.ts
```diff
@@ -0,0 +1,87 @@
...
+      repositoryName: 'data-prepper',
+    }).applyRemovalPolicy(removalPolicy);
+
+    new CfnPublicRepository(stack, 'logstash-oss-with-opensearch-output-plugin', {
+      repositoryName: 'logstash-oss-with-opensearch-output-plugin',
+    }).applyRemovalPolicy(removalPolicy);
+
+    new CfnPublicRepository(stack, 'opensearch-dashboards', {
+      repositoryName: 'opensearch-dashboards',
+    }).applyRemovalPolicy(removalPolicy);
```
gaiksaya, line 85: Convert to configuration, as ECR repo names depends on user and not everyone needs to follow the same.

bin/ci-stack.ts
abhinavGupta16: Changed the variable name to `defaultEnv`

README.md
```diff
@@ -35,12 +35,31 @@ OpenSearch Continuous Integration is an open source CI system for OpenSearch and
...
+      envName: 'dev',
+      env: {
+          // public ECR repositories can only be created in us-east-1
+          // https://github.com/aws/aws-cli/issues/5917#issuecomment-775564831
+          region: 'us-east-1',
+      },
+      repositories: [
+        'opensearch',
+      ],
+    }); ```
```
gaiksaya, line 60: Convert them into words instead of copying the codebase

lib/deploy-aws-assets.ts
```diff
@@ -0,0 +1,29 @@
...
+}
+
+export class DeployAwsAssets extends Stack {
+  constructor(scope: Construct, id: string, props: DeployAssetsProps) {
+    super(scope, id, props);
+    const deployECRParameter = `${props?.deployECR ?? this.node.tryGetContext('deployEcr')}`;
+
+    new CfnParameter(this, 'deployEcr', {
+      description: 'Should ECR repositories + roles be created',
+      default: deployECRParameter === 'true',
```
gaiksaya, line 18: I think default should be false. These are optional AWS assets that is not essential to jenkins deployment

test/ci-ecr-stack.test.ts
```diff
@@ -0,0 +1,168 @@
...
+  const deployAwsAssetsStack = new DeployAwsAssets(app, 'TestStack', deployAssetsProps);
+
+  const ecrStack = new CiEcrStack(deployAwsAssetsStack, 'ecrStack', deployAssetsProps);
+
+  // THEN
+  expect(ecrStack).to(countResources('AWS::ECR::PublicRepository', 0));
+  expect(deployAwsAssetsStack).to(countResources('AWS::ECR::PublicRepository', 0));
+  expect(ecrStack).to(countResources('AWS::IAM::Role', 1));
+  expect(ecrStack).to(countResources('AWS::IAM::ManagedPolicy', 1));
+  expect(deployAwsAssetsStack).to(countResources('AWS::CloudFormation::Stack', 1));
```
gaiksaya, line 70: I don't think this test case is required.

gaiksaya (changes requested):

lib/compute/jenkins-plugins.ts
```diff
@@ -40,6 +40,7 @@ export class JenkinsPlugins {
       'display-url-api',
       'docker-commons',
       'docker-custom-build-environment',
+      'docker-workflow',
```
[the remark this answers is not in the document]
abhinavGupta16: yes. I had this missed in the previous docker PR

lib/deploy-aws-assets.ts
abhinavGupta16: The default is `false`. This will evaluate to `false` if the value is anything other than `true`

test/ci-ecr-stack.test.ts
abhinavGupta16: It is. This test case verifies the behavior of the code if the repositories are not specified.

abhinavGupta16: It is. This test case verifies the behavior of the code if the repositories are not specified.

README.md
abhinavGupta16: The point describe how to deploy the assets as in previous points and comments describe the properties. Is there something specific you are looking for?

lib/deploy-aws-assets.ts
gaiksaya: Why not take it as cfn inputs that are passed as parameters like 'failOnCloudInitError' instead of getting it from context? Default it to false and if the customer needs they can add the value in props like this: https://github.com/opensearch-project/opensearch-ci/blob/main/lib/ci-stack.ts#L33-L35

README.md
gaiksaya: We already have this code in our codebase, you can just point at it and explain what you are trying to say

lib/deploy-aws-assets.ts
abhinavGupta16: It rather mimics the behavior of `useSsl` and `runWithOidc`. https://github.com/opensearch-project/opensearch-ci/blob/main/lib/ci-stack.ts#L29-L31

Doing it like this gives the user to control the behavior from the command line rather than modifying the code to make the changes. This can always be passed as a parameter too if someone wants to do that as here - https://github.com/opensearch-project/opensearch-ci/pull/77/files/f3027a2459b8bd3ac3c23b30d1e341dba7fadc99#diff-d638b1fcda588e6ca1f48230caf1d17595dccafad1e580bcd937a15150343471R11

This line however is only to record the changes to the parameter to cloud formation and not take input as https://github.com/opensearch-project/opensearch-ci/blob/main/lib/ci-stack.ts#L67-L70

README.md
abhinavGupta16: The reason I added this here in the `readme` as it is done for [here](https://github.com/opensearch-project/opensearch-ci/blob/main/README.md?plain=1#L39) is because defines every property that deploys the stack which should be a part of the readme point which talks about deploying `DeployAwsAssets` stack. The parameters in `ci-stack.ts` can then change on how user wants to configure it.

gaiksaya: Feels weird to see the same code base in readme. Customer can go and see for themselves. 
Can you rethink and rephrase this? 
Thanks!

lib/deploy-aws-assets.ts
gaiksaya: Was wondering if we really need it to provide it as cli param. Keeping it default as false would work as well. Anyway its a nit, we can refactor later if this doesn't work out.

README.md
abhinavGupta16: Updated the readme

gaiksaya (approved):

bbarani (approved):

PR #78, opened 2022-02-24
Add read-only role for anonymous users

gaiksaya:

### Description
For external users/contributors who are unable to login via OIDC, this change added a readonly permission for `anonymous` users.
 
### Issues Resolved
 
### Check List

### DIFF
--- a/lib/compute/jenkins-main-node.ts
+++ b/lib/compute/jenkins-main-node.ts
@@ -353,6 +353,7 @@ export class JenkinsMainNode {
         ? 'amazon-linux-extras install epel -y && yum -y install xmlstarlet'
         : 'echo not installing xmlstarlet as not running with OIDC'),
 
+      // Enabling Role Based Authentication with admin and read-only role
       InitCommand.shellCommand(oidcFederateProps.runWithOidc
         ? 'xmlstarlet ed -L -d /hudson/authorizationStrategy'
           + ' -s /hudson -t elem -n authorizationStrategy -v " "'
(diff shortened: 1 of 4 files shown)


lib/compute/jenkins-main-node.ts
```diff
@@ -366,17 +367,29 @@ export class JenkinsMainNode {
...
           + ' -i /hudson/authorizationStrategy/roleMap[3] -t attr -n "type" -v "slaveRolesRoles"'
+          + ' -s /hudson/authorizationStrategy/roleMap[2] -t elem -n role -v " "'
+          + ' -i /hudson/authorizationStrategy/roleMap[2]/role[2] -t attr -n "name" -v "read-only"'
+          + ' -i /hudson/authorizationStrategy/roleMap[2]/role[2] -t attr -n "pattern" -v ".*"'
+          + ' -s /hudson/authorizationStrategy/roleMap[2]/role[2] -t elem -n permissions -v " "'
+          // eslint-disable-next-line max-len
+          + `${JenkinsMainNodeConfig.readOnlyRolePermissions().map((e) => ` -s /hudson/authorizationStrategy/roleMap[2]/role[2]/permissions -t elem -n "permission" -v ${e}`).join(' ')}`
+          + ' -s /hudson/authorizationStrategy/roleMap[2]/role[2] -t elem -n "assignedSIDs" -v " "'
+          + ' -s /hudson/authorizationStrategy/roleMap[2]/role[2]/assignedSIDs -t elem -n "sid" -v "anonymous"'
           + ' /var/lib/jenkins/config.xml'
```
abhinavGupta16, line 384: Given that the shell commands don't wait for previous commands to finish, this can potentially cause a problem if the config is being reloaded while the next shell command edits the file.

The best approach in my opinion here would be to try to reload the config only once after the file is edited (We'll need to test this manually first) and maybe even add sleep if required (given file editing is a slow operation).

abhinavGupta16 (changes requested):

lib/compute/jenkins-main-node.ts
gaiksaya: The config is reloaded in the next line. I tested this code and can see the changes as expected without any interruption

gaiksaya: Added sleep and tested. Working as expected.

abhinavGupta16 (approved):

zelinh (approved):
Thanks

Issue #79, opened 2022-02-25
Add settings so opensearch-ci can run docker


abhinavGupta16:
### Is your feature request related to a problem? Please describe

Currently, no jenkins job will run on jenkins since docker and `withAws` is not installed.

### Describe the solution you'd like

Make changes so jenkins jobs with docker agent nodes are able to run.

PR #80, opened 2022-02-25
update jenkins deployment to use docker on agent nodes

abhinavGupta16:
Signed-off-by: Abhinav Gupta <abhng@amazon.com>

### Description
ci can now run docker agent nodes and withAws commands
 
### Issues Resolved
https://github.com/opensearch-project/opensearch-ci/issues/79
 
### Check List

### DIFF
--- a/lib/compute/agent-nodes.ts
+++ b/lib/compute/agent-nodes.ts
@@ -35,7 +35,8 @@ export class AgentNode {
 
   public readonly InitScript: string = 'sudo mkdir -p /var/jenkins/ && sudo chown -R ec2-user:ec2-user /var/jenkins '
     + '&& sudo yum install -y java-1.8.0-openjdk cmake python3 python3-pip && sudo yum groupinstall -y \'Development Tools\' '
-    + '&& sudo ln -sfn `which pip3` /usr/bin/pip && pip3 install pipenv && sudo ln -s ~/.local/bin/pipenv /usr/local/bin';
+    + '&& sudo ln -sfn `which pip3` /usr/bin/pip && pip3 install pipenv && sudo ln -s ~/.local/bin/pipenv /usr/local/bin '
+    + '&& sudo amazon-linux-extras install -y docker && sudo service docker start && sudo systemctl enable docker && sudo chmod 666 /var/run/docker.sock ';
 
   constructor(stack: Stack) {
     const key = new KeyPair(stack, 'AgentNode-KeyPair', {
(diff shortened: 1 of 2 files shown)


lib/compute/agent-nodes.ts
```diff
@@ -35,7 +35,8 @@ export class AgentNode {
 
   public readonly InitScript: string = 'sudo mkdir -p /var/jenkins/ && sudo chown -R ec2-user:ec2-user /var/jenkins '
     + '&& sudo yum install -y java-1.8.0-openjdk cmake python3 python3-pip && sudo yum groupinstall -y \'Development Tools\' '
-    + '&& sudo ln -sfn `which pip3` /usr/bin/pip && pip3 install pipenv && sudo ln -s ~/.local/bin/pipenv /usr/local/bin';
+    + '&& sudo ln -sfn `which pip3` /usr/bin/pip && pip3 install pipenv && sudo ln -s ~/.local/bin/pipenv /usr/local/bin '
```
gaiksaya, line 38: Would it make sense to move this init script at config level so that it is separate for each agent node?
We started with common script because both were AL2 agent nodes but Ubuntu, windows, etc will be different. Can be part of another PR though.

gaiksaya (approved):

lib/compute/jenkins-plugins.ts
```diff
@@ -92,6 +92,7 @@ export class JenkinsPlugins {
       'ownership',
       'pam-auth',
       'parameterized-scheduler',
+      'pipeline-aws',
```
peterzhuamazon, line 95: Why is this needed to be added? What purpose?

peterzhuamazon: withAWS?

lib/compute/agent-nodes.ts
peterzhuamazon: We havent think about windows for now, but only using AL2 for LINUX Agent is enough, as AL2 can provision Ubuntu container for buildx purposes.

gaiksaya: `AL2 can provision Ubuntu container for buildx purposes` I didn't get that part? You mean we can pull ubuntu image on AL2?

abhinavGupta16: I tried that and it works

abhinavGupta16: I was able to pull the ubuntu docker image on AL2X64 and run the job

zelinh (approved):

PR #83, opened 2022-03-03
Lock plugin versions

### LINKED ISSUE #27: [Bug] Plugin versions should be locked in the cdk
From https://github.com/opensearch-project/opensearch-ci/pull/26

> .@gaiksaya
> Quick question about jenkins plugin abhinavGupta16 if we do not specify the version of these plugins, won't it install most recent version that can differ in functionality in someway? peterzhuamazon can weigh in here regarding plugins and how important their versioning is (or not important)
>  
> .@abhinavGupta16
> The script that ran to install plugins did not have versions for the plugins. This is a good point we should think about.
>  
> .@peterzhuamazon
> I dont see this as a problem as we want to update the plugins to the latest versions to patch all vulnerabilities.
> However, I also think maintaining a list of versions for stable usage is also beneficial.
> You can have 1 template for latest and 1 template for stable. What do you think about this?

gaiksaya:

### Description
Since we are modifying the config.xml as well as various other configurations that are Jenkins Plugins based, this PR locks the version so that a specific compatible version of plugin is installed on each deployment.
 
### Issues Resolved
resolves #27 
 
### Check List


abhinavGupta16 (approved):
Thanks for making this change

Issue #88, opened 2022-03-18
Load jenkins via jenkins.yaml


gaiksaya:
Load configuration via code using Jenkins Configuration as Code plugin that uses jenkins.yaml to load config. We need load the entire config via jenkins.yaml. This means we need to move current config changes made to config.xml to jenkins.yaml

- [x]   - OIDC config (also includes Role Based Auth config) : See #89 
- [ ]   - Setting up agent nodes

_Originally posted by @gaiksaya in https://github.com/opensearch-project/opensearch-ci/issues/85#issuecomment-1071507828_

Issue #85, opened 2022-03-10
Data retention for Jenkins


gaiksaya:
## Problem Statement
With current set up whenever there is a change with EC2 configuration or change in InitCommands, the stack replaces the EC2 with new one with new configurations. In initial stages it is fine because there is no relevant data. However, once in advance stage the replacement of EC2 will lead to data loss as EC2 is replaced.

## Possible Solutions
Currently looking at the below data that needs to be retained:
1. Jobs and job history: Stored in $JENKINS_HOME/jobs folder
1. Config.xml : Contains all the global config changes (eg: env variables, any global settings, etc)
1. More? 

_We are uploading all logs from (/var/log/jenkins/jenkins.log) to cloudwatch so no need to have that._

_I do not have concrete solution for config.xml because of the following concern:_
Config.xml stores all the global configurations in Jenkins including env variables, global settings, etc.
If we replace the newly deployed config.xml, we overwrite new changes made by new deployment.
If we keep the new one, we overwrite all the old changes made by the system.
Possible solution: 
* Do all the config changes via code? https://www.jenkins.io/doc/book/managing/casc/

#### 1. Mount it on /data: 

- Sync only the required files with newly deployed system. 
- We would also need a reverse sync to keep the data on multi-attach EBS up to date.

**Pros:**
   1. Avoids data corruption as data in one way writable

**Cons:**
1. Need to manage or come up with the cronjob/script/lambda function that runs this sync inside EC2
4. Reading the data while Jenkins or System is actively writing might lead to partial data
5. Depending on the frequency of the sync, if the instance is replaced in between, the data between the last sync and node replacement will be lost

   
#### 2 Mount directly on required path eg: $JENKINS_HOME/jobs

**Pros:**

1. No syncing between 2 drives required
3. No external scripts to manage
4. No time gap for data loss

**Cons:**

1. Cloudformation first brings up new instance and then deletes the old one. With the old instance still writing to the file system, mounting the drive to new one can lead to data corruption and overwriting the old data.

#### 1. Multi-attach Elastic Block Storage volumes:
A multi-attach EBS volume that can be attached to the EC2 instance. 

Pros:
1. EBS is a good option for storage with high performance, low latency
1. Easy to mount and attach to EC2 during boot up time
1. Easy data backup and restoration – via snapshots that can be taken at hourly intervals
1. EBS ensures all your data is well protected
1. Multi-attach gives us flexibility to attach the same volume to terminating and initializing instances

Cons:
[...]

PR #89, opened 2022-03-18
[OIDC] Load configuration via jenkins.yaml

gaiksaya:

### Description
Please see issue #85 for more details. We want to load the configuration via source coded config file so that the redeployment of the EC2 instance does not result in loss of config change.  This will also ensure that we make all changes to configuration only via code and not via UI.
The baseYaml remains the same and as per the features, a new file is created and loaded in jenkins.
I also think we should increment the version of this package as this is a breaking change.
Note: Until #88  is fixed, this change breaks agent node configuration.
 
### Issues Resolved
Part of https://github.com/opensearch-project/opensearch-ci/issues/88
 
### Check List


lib/compute/jenkins-main-node.ts
```diff
@@ -191,6 +198,8 @@ export class JenkinsMainNode {
       InitPackage.yum('java-1.8.0-openjdk-devel'),
       InitPackage.yum('openssl'),
       InitPackage.yum('mod_ssl'),
+      // eslint-disable-next-line max-len
+      InitCommand.shellCommand('sudo wget -nv https://github.com/mikefarah/yq/releases/download/v4.22.1/yq_linux_amd64 -O /usr/bin/yq && sudo chmod +x /usr/bin/yq'),
```
abhinavGupta16, line 202: That is this repository used for?

lib/compute/jenkins-main-node.ts
```diff
@@ -368,71 +377,31 @@ export class JenkinsMainNode {
...
-        + ' xmlstarlet ed -L -d "/hudson/securityRealm"'
-        + ' -s /hudson -t elem -n securityRealm -v " "'
-        + ' -i //securityRealm -t attr -n "class" -v "org.jenkinsci.plugins.oic.OicSecurityRealm"'
-        + ' -i //securityRealm -t attr -n "plugin" -v "oic-auth@1.8"'
+        + ' varkeys=`echo $var | yq \'keys\' | cut -d "-" -f2 | cut -d " " -f2` &&'
         // eslint-disable-next-line max-len
-        + `${JenkinsMainNodeConfig.oidcConfigFields().map((e) => ` -s /hudson/securityRealm -t elem -n ${e[0]} -v ${e[1] === 'replace' ? `"$(echo $var | jq -r ".${e[0]}")"` : `"${e[1]}"`}`).join(' ')}`
-        + ' /var/lib/jenkins/config.xml'
-        + ` && ${RELOAD_CONFIG}`
+        + ' for i in $varkeys; do newvalue=`echo $var | yq .$i` && myenv=$newvalue i=$i yq -i \'.jenkins.securityRealm.oic.[env(i)]=env(myenv)\' /var/lib/jenkins/jenkins.yaml ; done'
```
abhinavGupta16, line 387: Shouldn't we parse the yml file instead? How is this different from xmlstarlet?

.gitignore
```diff
@@ -3,6 +3,8 @@
 !jest.config.js
 *.d.ts
 node_modules
+resources/jenkins.yaml
```
abhinavGupta16, line 6: Can you add a comment here explaining why are we ignoring this file?

abhinavGupta16 (changes requested):

lib/compute/jenkins-main-node.ts
gaiksaya: You cannot add aws secrets values in typescript. I am getting values and replacing them in already loaded yaml on the server

lib/compute/jenkins-main-node.ts
gaiksaya: Please refer the comment above. we are not adding any config via this command, Just replacing with user provided values

.gitignore
gaiksaya: will do

lib/compute/jenkins-main-node.ts
abhinavGupta16: This is exactly what we can do with xml as well right?

test/data/jenkins.yaml
```diff
@@ -0,0 +1,253 @@
...
+    tokenGenerationOnCreationEnabled: false
+    usageStatisticsEnabled: true
+  copyartifact:
+    mode: PRODUCTION
+  globalJobDslSecurityConfiguration:
+    useScriptSecurity: true
+  sSHD:
+    port: -1
+unclassified:
+  ansiColorBuildWrapper:
```
prudhvigodithi, line 84: we can remove this ansiColorBuildWrapper as by default this be coming default from ANSI plugin.

lib/compute/jenkins-main-node.ts
abhinavGupta16: Can we rather extract the values in EC2, return them and use them in code?

.gitignore
abhinavGupta16: thanks

lib/compute/jenkins-main-node.ts
gaiksaya: Nope, that will over complicate this unnecessarily. Plus from security point of view it is better that the secrets stay on server than written to local file.

lib/compute/jenkins-main-node.ts
gaiksaya: We are adding configuration via code (yaml in this case). You can check linked issues for details why we reached this decision. Also editing, maintaining and manipulating yaml is much easier than xml. Now on, every user does not have to learn xmlstartlet to add code to xml. It was a very messy process and changes in hudson libraries and config might have affected us. Right now, yaml you add the config you want in yaml and yaml takes care of it to add it to config. xml.

TL;DR:
Easy to maintain, codeify and update the config via code. (It was like we were editing bytecodes by editing config.xml which will have serious implications and unknown errors.)

.gitignore
gaiksaya: Added

test/data/jenkins.yaml
gaiksaya: Removed.

lib/compute/jenkins-main-node.ts
abhinavGupta16: Do we have to learn `yq` now?

lib/compute/jenkins-main-node.ts
abhinavGupta16: The secret would still be written on the file on the server, just like now.

abhinavGupta16 (approved):

lib/compute/jenkins-main-node.ts
gaiksaya: Just for replacing any values in jenkins.yaml on the jenkins server directly during installation.

lib/compute/jenkins-main-node.ts
abhinavGupta16: discussed offline

PR #90, opened 2022-03-23
[Agent Nodes] Load jenkins config via yaml

### LINKED ISSUE #88: Load jenkins via jenkins.yaml

gaiksaya:

### Description
This change adds the functionality to load agent nodes configuration via jenkins.yaml

### Testing
Tested by deploying the stack using all combinations(ssl/oidc: true/false) to make sure none of the other functionality breaks.
 
### Issues Resolved
Resolves #88 
 
### Check List

### DIFF
--- a/lib/compute/agent-node-config.ts
+++ b/lib/compute/agent-node-config.ts
@@ -6,39 +6,108 @@
  * compatible open source license.
  */
 
-import {
-  AmazonLinuxCpuType, AmazonLinuxGeneration, InstanceClass, InstanceSize, InstanceType, MachineImage,
-} from '@aws-cdk/aws-ec2';
-import { Stack } from '@aws-cdk/core';
-import { AgentNodeConfig } from './agent-nodes';
+import { CfnInstanceProfile, ServicePrincipal, Role } from '@aws-cdk/aws-iam';
+import { Fn, Stack, Tags } from '@aws-cdk/core';
+import { KeyPair } from 'cdk-ec2-key-pair';
+import { readFileSync } from 'fs';
+import { load } from 'js-yaml';
+import { JenkinsMainNode } from './jenkins-main-node';
 
-export class CloudAgentNodeConfig {
-  readonly AL2_X64: AgentNodeConfig;
+export interface AgentNodeProps {
+  amiId: string;
+  instanceType: string;
+  workerLabelString: string;
+  remoteUser: string;
+  initScript: string
+}
 
-  readonly AL2_ARM64: AgentNodeConfig;
+export interface AgentNodeNetworkProps {
+  readonly agentNodeSecurityGroup: string;
+  readonly subnetId: string;
+}
+
+export class AgentNodeConfig {
+  public readonly AgentNodeInstanceProfileArn: string;
+
+  public readonly STACKREGION: string;
+
+  public readonly SSHEC2KeySecretId: string;
 
   constructor(stack: Stack) {
-    this.AL2_X64 = {
-      ec2CloudName: 'AL2-X64',
-      instanceType: InstanceType.of(InstanceClass.C5, InstanceSize.XLARGE4).toString(),
-      workerLabelString: 'AL2-X64',
-      numberOfExecutors: '2',
-      remoteUser: 'ec2-user',
-      amiId: MachineImage.latestAmazonLinux({
-        generation: AmazonLinuxGeneration.AMAZON_LINUX_2,
-        cpuType: AmazonLinuxCpuType.X86_64,
-      }).getImage(stack).imageId.toString(),
-    };
-    this.AL2_ARM64 = {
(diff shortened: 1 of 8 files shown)


lib/compute/oidc-config.ts
```diff
@@ -55,8 +52,8 @@ export class OidcConfig {
       'Job/Read',
     ];
 
-    public static addOidcConfigToJenkinsYaml(newYamlPath: string, admins?: string[]): void {
-      const jenkinsYaml: any = load(readFileSync(JenkinsMainNode.BASE_JENKINS_YAML_PATH, 'utf-8'));
+    public static addOidcConfigToJenkinsYaml(yamlObject: any, admins?: string[]): any {
```
abhinavGupta16, line 55: Would yamlObject be a Map?

gaiksaya: Nope it is an object https://github.com/opensearch-project/opensearch-ci/pull/90/files#diff-4c90022915859e4f2bab2c590f837062e3eb5a708c33a6d75d4ee20a149228f4R53

lib/compute/agent-node-config.ts
```diff
@@ -6,39 +6,108 @@
...
+      hostKeyVerificationStrategy: 'OFF',
+      iamInstanceProfile: this.AgentNodeInstanceProfileArn,
+      idleTerminationMinutes: '30',
+      initScript: config.initScript,
+      labelString: config.workerLabelString,
+      launchTimeoutStr: '300',
+      maxTotalUses: -1,
+      minimumNumberOfInstances: 0,
+      minimumNumberOfSpareInstances: 1,
+      mode: 'EXCLUSIVE',
```
peterzhuamazon, line 95: What is this exclusive means here?

peterzhuamazon (commented):
One thing I want to call out is there seems like a toggle in agent node, to auto delete it if not able to start up in x minutes.

I have seen a few agent node in past days that stuck for weeks and I have to manually delete them.

This is quite common if there is internet issues causing the `yum update` or similar to fail in the middle of start.

Thanks.

lib/compute/agent-node-config.ts
gaiksaya: Looks like node will only be used when that node label is specified: https://github.com/jenkinsci/ec2-plugin/blob/ec2-1.38/src/main/java/hudson/plugins/ec2/SlaveTemplate.java#L1172-L1176
Not sure but most probably it maps to this setting:
image

gaiksaya:
> One thing I want to call out is there seems like a toggle in agent node, to auto delete it if not able to start up in x minutes.
> 
> I have seen a few agent node in past days that stuck for weeks and I have to manually delete them.
> 
> This is quite common if there is internet issues causing the `yum update` or similar to fail in the middle of start.
> 
> Thanks.

Do you the settings that handles this behavior?

lib/compute/agent-node-config.ts
peterzhuamazon: Oh you mean this, this is good and should be this way.

peterzhuamazon:
> > One thing I want to call out is there seems like a toggle in agent node, to auto delete it if not able to start up in x minutes.
> > I have seen a few agent node in past days that stuck for weeks and I have to manually delete them.
> > This is quite common if there is internet issues causing the `yum update` or similar to fail in the middle of start.
> > Thanks.
> 
> Do you the settings that handles this behavior?

I dont have that setting in handy, but you should find it somewhere in the GUI in cloud setup.

gaiksaya:
> I dont have that setting in handy, but you should find it somewhere in the GUI in cloud setup.

I see `Launch Timeout in seconds` which is set to 300 for now.

lib/compute/oidc-config.ts
abhinavGupta16: Can it be a `Map` object?

peterzhuamazon:
> > I dont have that setting in handy, but you should find it somewhere in the GUI in cloud setup.
> 
> I see `Launch Timeout in seconds` which is set to 300 for now.

which means it didnt take into effect, but we can improve it later on.

peterzhuamazon (approved):

lib/compute/oidc-config.ts
gaiksaya: js-yaml load function returns `unknown` hence used `any` to perform operations on the same. 
We can enhance later is required.

PR #99, opened 2022-04-13
Support data retention for jenkins jobs

### LINKED ISSUE #85: Data retention for Jenkins

gaiksaya:
### Description
This PR adds support to enable data retention for jenkins. The only data that is retained on EFS (external storage) is jenkins jobs and its build history. `var/lib/jenkins/jobs`
For jenkins system configuration we use configuration as code where as logs logged in cloudwatch `/var/log/jenkins/jenkins.log`
 
### Issues Resolved
closes #85 
 
### Check List

### DIFF
--- a/lib/compute/jenkins-main-node.ts
+++ b/lib/compute/jenkins-main-node.ts
@@ -24,7 +24,9 @@ import {
   SubnetType,
   Vpc,
 } from '@aws-cdk/aws-ec2';
-import { Effect, IManagedPolicy, ManagedPolicy, PolicyStatement, Role, ServicePrincipal, } from '@aws-cdk/aws-iam';
+import {
+  Effect, IManagedPolicy, ManagedPolicy, PolicyStatement, Role, ServicePrincipal,
+} from '@aws-cdk/aws-iam';
 import { Metric, Unit } from '@aws-cdk/aws-cloudwatch';
 import { join } from 'path';
 import { dump } from 'js-yaml';
@@ -77,7 +79,7 @@ export class JenkinsMainNode {
 
   static readonly JENKINS_DEFAULT_ID_PASS_PATH: String = '/var/lib/jenkins/secrets/myIdPassDefault';
 
-  public readonly EFS_ID: string;
+  private readonly EFS_ID: string;
 
   public readonly ec2Instance: Instance;
 
@@ -87,9 +89,7 @@ export class JenkinsMainNode {
     foundJenkinsProcessCount: Metric
   }
 
-  constructor(stack: Stack,
-    props: JenkinsMainNodeProps,
-    agentNode: AgentNodeProps[]) {
+  constructor(stack: Stack, props: JenkinsMainNodeProps, agentNode: AgentNodeProps[]) {
     this.ec2InstanceMetrics = {
       cpuTime: new Metric({
         metricName: 'procstat_cpu_usage',
--- a/lib/compute/jenkins-main-node.ts
+++ b/lib/compute/jenkins-main-node.ts
@@ -6,18 +6,30 @@
  * compatible open source license.
  */
 
-import { Duration, Stack } from '@aws-cdk/core';
+import { Duration, RemovalPolicy, Stack } from '@aws-cdk/core';
 import {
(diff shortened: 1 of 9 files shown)


prudhvigodithi (approved):

lib/compute/jenkins-main-node.ts
```diff
@@ -60,6 +77,8 @@ export class JenkinsMainNode {
 
   static readonly JENKINS_DEFAULT_ID_PASS_PATH: String = '/var/lib/jenkins/secrets/myIdPassDefault';
 
+  public readonly EFS_ID: string;
```
abhinavGupta16, line 80: Do we need a class level variable public variable for `EFS_ID`? Let's declare and use it locally.

gaiksaya: Its conditional statement so need it at class level

abhinavGupta16: We can declare it outside the condition and use it

gaiksaya: Since config element is static it throws an error. Making it private that will prevent outside access to this variable.
Thanks for catching this

abhinavGupta16 (approved):

PR #126, opened 2022-05-27
Added new jenkins agents and logging bug-fix

### LINKED ISSUE #125: [Bug]: Hardcoded log-stream causing intermittent stack bring up failure
### Describe the bug

Whenever there is a change that causes the jenkins main node to be replaced we see intermittent failure during instance initialisation. 

**Cause**:
This is due to hardcoded log-stream value [here](https://github.com/opensearch-project/opensearch-ci/blob/main/resources/docker-compose.yml#L28). 
Whenever a change causes a new jenkins main node to be brought up, the moment the new main node is brought up and docker-compose is run, the new container tries to write its logs to an existing log-stream which is already in use by the previous jenkins main node container which is yet to be brought down. 
As per https://docs.docker.com/config/containers/logging/awslogs/#awslogs-stream two containers writing to the same log-stream is not a good practise and results in degraded performance. This is the reason we see intermittent failures during new instance bring up. 

### To reproduce

Bring up a new CI stack, make a change that will potentially result in jenkins main node replacement and redeploy the stack. 

### Expected behavior

_No response_

### Host / Environment

_No response_

### Additional context

_No response_

### Relevant log output

```shell
May 26 15:03:03 ip-10-0-145-33 dockerd: time="2022-05-26T15:03:03.838804152Z" level=error msg="Failed to put log events" errorCode=InvalidSequenceTokenException logGroupName=JenkinsMainNode/jenkins.log logStreamName=jenkins.log message="The given sequenceToken is invalid. The next expected sequenceToken is: 49629487109106147829027718688550912310461956053603975282" origError="<nil>"
May 26 15:03:04 ip-10-0-145-33 systemd: Unmounted /run/docker/netns/f01cf084571d.
```

rishabh6788:
Signed-off-by: Rishabh Singh <sngri@amazon.com>

### Description
The PR contains below updates:
1. Adds 2 new jenkins agent configuration. 
2. Adds missing plugin `pipeline-utility-steps` that is required for build jobs. 
3. Removes hardcoded `log-stream` values in docker-compose configuration as it was causing multiple containers writing to same stream, therefore, resulting in errors. See #125 

### Issues Resolved
#125 


lib/compute/agent-nodes.ts
```diff
@@ -31,6 +35,13 @@ export class AgentNodes {
         + ' sudo yum install -y java-1.8.0-openjdk cmake python3 python3-pip python3-devel && sudo yum groupinstall -y "Development Tools" &&'
         + ' sudo ln -sfn `which pip3` /usr/bin/pip && pip3 install pipenv && sudo ln -s ~/.local/bin/pipenv /usr/local/bin',
       };
+      this.AL2_X64_DOCKER_1 = {
+        workerLabelString: 'Jenkins-Agent-al2-x64-c54xlarge-Docker-Host',
+        instanceType: 'C54xlarge',
+        remoteUser: 'ec2-user',
+        amiId: 'ami-00a07e55fcad01043',
+        initScript: 'sudo yum update -y || sudo yum update -y',
```
prudhvigodithi, line 43: @rishabh6788  We should be installing `docker` as well

lib/compute/agent-nodes.ts
```diff
@@ -43,5 +54,12 @@ export class AgentNodes {
         + ' sudo yum install -y java-1.8.0-openjdk cmake python3 python3-pip python3-devel && sudo yum groupinstall -y "Development Tools" &&'
         + ' sudo ln -sfn `which pip3` /usr/bin/pip && pip3 install pipenv && sudo ln -s ~/.local/bin/pipenv /usr/local/bin',
       };
+      this.AL2_ARM64_DOCKER_1 = {
+        workerLabelString: 'Jenkins-Agent-al2-arm64-c6g4xlarge-Docker-Host',
+        instanceType: 'C6g4xlarge',
+        remoteUser: 'ec2-user',
+        amiId: 'ami-020c52efb1a60f1ae',
+        initScript: 'sudo yum update -y || sudo yum update -y',
```
prudhvigodithi, line 62: Same docker should also be installed

prudhvigodithi:
@rishabh6788 please change the following as well inside the compose file, this can go with this PR as this is minimal, thoughts @rishabh6788 ?
```
    deploy:
      resources:
        limits:
          cpus: '4'
          memory: 10000m
        reservations:
          cpus: '4'
          memory: 10000m
```

lib/compute/agent-nodes.ts
rishabh6788: Docker is already installed on the ami.

rishabh6788:
> I did try deploying this change twice and both the times the docker-compose bring up failed and terminated the instance. Unfortunately both the time the system did not allow me to ssh and see the logs. So for safer side I kept the change as is since it is working. 
Will debug this in the next iteration. @prudhvigodithi

prudhvigodithi:
> I did try deploying this change twice and both the times the docker-compose bring up failed and terminated the instance. Unfortunately both the time the system did not allow me to ssh and see the logs. So for safer side I kept the change as is since it is working.
> Will debug this in the next iteration. @prudhvigodithi

sure @rishabh6788

prudhvigodithi:
I remember that log group was added for an audit purpose, I'm approving, but we should be revisiting the log group settings.
@gaiksaya @bbarani @peterzhuamazon

prudhvigodithi (approved):

rishabh6788:
@prudhvigodithi Sure. This change wouldn't remove the log group. This change will make sure that every time a new jenkins container is deployed it doesn't write to the same log-stream, which is a bad practise, instead, it will create a new log-stream in the same log group and the name of the log-stream would be container id by default.

PR #147, opened 2022-06-22
Main/Agent Node new setups on tag and executors

### LINKED ISSUE #146: [Enhancement] Let each agent node have its specific nametag/executors in EC2
As of now all of the agent node in ec2 is consuming the same name tag.
image

We want to have different name tags based on label such like this:
image

This is controlled in GUI by this:
image

Also, the Single-Host and Docker-Builder requires to have only 1 executor.
While the Docker-Host one can up to 6 or 8.

Will change those as well here.

Thanks.

peterzhuamazon:

### Description
Main/Agent Node new setups on tag and executors

All changes:
1. Replace complicated setup with existing AMIs, simplify agent nodes.
2. Set agent node name tag in EC2 using their labels.
3. Add abilities to separate number of executors based on different agent node setups, formalize tag naming.
4. Set 4 executors for main node, set main-node tag, set only to run if tag is matched to avoid wrong scheduling.
5. Remove DCO check workflow as the app is already taken over the task.
6. Please advice if there is any test cases required to be added here.

Thanks.

### Issues Resolved
#146


lib/compute/agent-nodes.ts
```diff
@@ -16,59 +16,55 @@ export class AgentNodes {
...
-    readonly AL2_ARM64_DOCKER_1: AgentNodeProps;
+    readonly AL2_ARM64_DOCKER_HOST: AgentNodeProps;
 
-    readonly UBUNTU_X64_DOCKER: AgentNodeProps;
+    readonly UBUNTU_X64_DOCKER_BUILDER: AgentNodeProps;
 
     constructor(stack: Stack) {
       this.AL2_X64 = {
-        workerLabelString: 'AL2-X64',
+        workerLabelString: 'Jenkins-Agent-al2-x64-c54xlarge-Single-Host',
```
gaiksaya, line 29: Everything is jenkins here. How about we remove `Jenkins` form labels.

lib/compute/agent-nodes.ts
```diff
@@ -16,59 +16,55 @@ export class AgentNodes {
...
+        numExecutors: 1,
+        amiId: 'ami-00a07e55fcad01043',
+        initScript: 'sudo yum update -y || sleep 10 && sudo yum update -y',
       };
-      this.AL2_X64_DOCKER_1 = {
+      this.AL2_X64_DOCKER_HOST = {
         workerLabelString: 'Jenkins-Agent-al2-x64-c54xlarge-Docker-Host',
         instanceType: 'C54xlarge',
         remoteUser: 'ec2-user',
+        numExecutors: 8,
```
gaiksaya, line 40: Won't 8 be too much? How about 4 to start with?

lib/compute/agent-nodes.ts
```diff
@@ -16,59 +16,55 @@ export class AgentNodes {
...
+        numExecutors: 1,
+        amiId: 'ami-020c52efb1a60f1ae',
+        initScript: 'sudo yum update -y || sleep 10 && sudo yum update -y',
       };
-      this.AL2_ARM64_DOCKER_1 = {
+      this.AL2_ARM64_DOCKER_HOST = {
         workerLabelString: 'Jenkins-Agent-al2-arm64-c6g4xlarge-Docker-Host',
         instanceType: 'C6g4xlarge',
         remoteUser: 'ec2-user',
+        numExecutors: 8,
```
gaiksaya, line 56: Same as above, how about 4?

lib/compute/agent-nodes.ts
rishabh6788: ideally it should be half the number of cores.

lib/compute/agent-nodes.ts
rishabh6788: Our jobs look for this label with `Jenkins` keyword in it. If we are removing here then let us make sure we update our job definitions as well.

peterzhuamazon: There are a lot of existing pipeline files using this format for Docker-Host, and I am more support of adding `Jenkins-Agent` because that is clear to tell you this is an agent node, not a container.

lib/compute/agent-nodes.ts
peterzhuamazon: 8 is not too much we have 16 core here and each use 2 core is fine, we have been using this exact same setup in jenki-jenki for nearly a year.

lib/compute/agent-nodes.ts
peterzhuamazon: Same as above.

rishabh6788 (approved):

lib/compute/agent-nodes.ts
```diff
@@ -16,59 +16,55 @@ export class AgentNodes {
...
     constructor(stack: Stack) {
       this.AL2_X64 = {
-        workerLabelString: 'AL2-X64',
+        workerLabelString: 'Jenkins-Agent-al2-x64-c54xlarge-Single-Host',
         instanceType: 'C54xlarge',
         remoteUser: 'ec2-user',
-        amiId: MachineImage.latestAmazonLinux({
-          generation: AmazonLinuxGeneration.AMAZON_LINUX_2,
-          cpuType: AmazonLinuxCpuType.X86_64,
-        }).getImage(stack).imageId.toString(),
```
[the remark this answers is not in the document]
gaiksaya: I prefer to clean the code rather than accumulating technical debt.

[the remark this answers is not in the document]
peterzhuamazon: Already removed.

PR #149, opened 2022-06-24
Add Ubuntu Single Host for Gradle Check

peterzhuamazon:

### Description
Add Ubuntu Single Host for Gradle Check

### Issues Resolved
https://github.com/opensearch-project/opensearch-build/issues/851


peterzhuamazon:
@gaiksaya @prudhvigodithi @rishabh6788 

This PR include these changes for gradle check migration:
1. Add a standalone Ubuntu Single Host with `docker-compose` installed.
2. Add JDK tools 8 / 11 / 17 to install directly on Agent Node, as gradle check requires all 3 JDKs to present, and have issues running on Docker, as it will start containers on the agent for testing, thus requires Jenkins to provide the JDK installation through AdoptOpenJDK plugin.
3. Tweak AL2 agent node init script to avoid consistent yum transaction error, as it caused by old cache of yum. If this still not resolve completely I would force a --skip-broken later. In current test it runs pretty well with this new init.

Thanks.

prudhvigodithi (approved):

PR #178, opened 2022-08-02
Add ability to assume multiple roles

gaiksaya:

### Description
A single agent node role might need to assume multiple roles across different accounts. This PR adds that functionality.
Also removes policy name that causes conflict during multiple deployments.

### Issues Resolved

### DIFF
--- a/lib/compute/agent-node-config.ts
+++ b/lib/compute/agent-node-config.ts
@@ -42,7 +42,7 @@ export class AgentNodeConfig {
 
    public readonly SSHEC2KeySecretId: string;
 
-   constructor(stack: Stack, assumeRole: string) {
+   constructor(stack: Stack, assumeRole?: string[]) {
      this.STACKREGION = stack.region;
      this.ACCOUNT = stack.account;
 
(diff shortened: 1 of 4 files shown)


lib/compute/agent-node-config.ts
```diff
@@ -107,11 +106,11 @@ export class AgentNodeConfig {
      );
 
      /* eslint-disable eqeqeq */
-     if (assumeRole.toString() !== 'undefined') {
+     if (assumeRole) {
```
prudhvigodithi, line 109: The `assumeRole` is not a boolean, have you tested this @gaiksaya ?

gaiksaya: Yes, works as expected. Also the tests cover this scenario here in this file `test/compute/agent-node-config.test.ts`

gaiksaya: Also if the resource string array checks for the validity of the resource we are adding, it cannot be blank or random values has to be a valid resource like arns or *

prudhvigodithi (approved):

rishabh6788 (approved):

PR #194, opened 2022-08-26
Increase Jenkins Main Node to c59xlarge instance

peterzhuamazon:

### Description
Increase Jenkins Main Node to c59xlarge instance

### Issues Resolved

### DIFF
--- a/resources/baseJenkins.yaml
+++ b/resources/baseJenkins.yaml
@@ -15,7 +15,7 @@ jenkins:
       disableSyntaxHighlighting: true
   mode: EXCLUSIVE
   myViewsTabBar: "standard"
-  numExecutors: 4
+  numExecutors: 10
   primaryView:
     all:
       name: "all"
--- a/lib/compute/jenkins-main-node.ts
+++ b/lib/compute/jenkins-main-node.ts
@@ -116,7 +116,7 @@ export class JenkinsMainNode {
       this.EFS_ID = efs.fileSystemId;
     }
     this.ec2Instance = new Instance(stack, 'MainNode', {
-      instanceType: InstanceType.of(InstanceClass.C5, InstanceSize.XLARGE4),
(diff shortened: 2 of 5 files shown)


resources/baseJenkins.yaml
```diff
@@ -15,7 +15,7 @@ jenkins:
       disableSyntaxHighlighting: true
   mode: EXCLUSIVE
   myViewsTabBar: "standard"
-  numExecutors: 4
+  numExecutors: 10
```
gaiksaya, line 18: Will increasing the executors cause the same problem as before even though we increased the size?
Also jenkins recommends to keep this value as 0 can we try that? It is not a good practice to keep executors on main node. Jenkins takes care of scheduling the jobs at the back end so executors are not actually needed

peterzhuamazon: We are still doing some scheduling on main node so we need some executors.
I have seen main node scheduling using the executors.

gaiksaya: As per the documentation it only executes the jobs and nothing else:
https://www.jenkins.io/doc/book/managing/nodes/

>An executor is a slot for execution of tasks; effectively, it is a thread in the agent. The number of executors on a node defines the number of concurrent tasks that can be executed on that node at one time. In other words, this determines the number of concurrent Pipeline stages that can execute on that node at one time.

Also see: https://www.jenkins.io/doc/book/security/controller-isolation/
Can we atleast try and test this theory? I don't think increasing this will make any difference but it should be 0 for security reasons.
would like you thoughts here @prudhvigodithi

peterzhuamazon: We can try on the staging setups.

prudhvigodithi: Definitely we cant go with zero as few jobs initial clone happens on master, then followed by the label inside the Jenkinsfile.

peterzhuamazon:
Close this for now as it is not needed.

PR #215, opened 2022-11-02
Add JENKINS_HOME_PATH as default JENKINS_HOME always returns /var/jenkins_home

peterzhuamazon:

### Description
Add JENKINS_HOME_PATH as default JENKINS_HOME always returns /var/jenkins_home.
This is a backend settings required to support windows gradle check.
Due to LINUX and Windows having different home directory for Jenkins, we need a variable to export the value.
The issue is, the default JENKINS_HOME variable is always returning `/var/jenkins_home` regardless of the platform.

By adding a node var for the agents it will dynamically display the root based on the platform remote fs.

Thanks.

### Issues Resolved
https://github.com/opensearch-project/opensearch-build/issues/2306

### DIFF
--- a/resources/baseJenkins.yaml
+++ b/resources/baseJenkins.yaml
@@ -175,8 +175,16 @@ tool:
       properties:
       - installSource:
           installers:
-          - zip:
-              url: "https://ci.opensearch.org/ci/dbc/tools/OpenJDK14U-jdk_x64_linux_hotspot_14.0.2_12.zip"
+          - command:
+              command: "JENKINS_HOME_LINUX=\"/var/jenkins\"\nJENKINS_HOME_WINDOWS=\"\
+                C:/Users/Administrator/jenkins\"\n \n\nif uname -s | grep -i NT; then\n\
+                \    echo windows agent\n    JENKINS_JDK14_PATH=$JENKINS_HOME_WINDOWS/tools/hudson.model.JDK/openjdk-14\n\
+                \    JDK14_ZIP=\"OpenJDK14U-jdk_x64_windows_hotspot_14.0.2_12.zip\"\
+                \n    \nelse\n     echo linux agent\n     JENKINS_JDK14_PATH=$JENKINS_HOME_LINUX/tools/hudson.model.JDK/openjdk-14\n\
+                \     JDK14_ZIP=\"OpenJDK14U-jdk_x64_linux_hotspot_14.0.2_12.zip\"\
+                \n     \nfi\n\n\nmkdir -p $JENKINS_JDK14_PATH\ncd $JENKINS_JDK14_PATH\n\
+                pwd\nrm -rf ./*\ncurl -sSL https://ci.opensearch.org/ci/dbc/tools/$JDK14_ZIP\
+                \ -o jdk14.zip\nunzip -q jdk14.zip\nrm jdk14.zip\n\n"
     - name: "openjdk-17"
       properties:
       - installSource:
--- a/lib/compute/agent-node-config.ts
+++ b/lib/compute/agent-node-config.ts
@@ -196,6 +196,18 @@ export class AgentNodeConfig {
       ],
       tenancy: 'Default',
       type: config.instanceType,
+      nodeProperties: [
+        {
+          envVars: {
+            env: [
+              {
+                key: 'JENKINS_HOME_PATH',
+                value: config.remoteFs,
+              },
+            ],
+          },
+        },
+      ],
       useEphemeralDevices: false,
     };
   }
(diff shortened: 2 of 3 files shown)


bbarani (approved):

resources/baseJenkins.yaml
```diff
@@ -175,8 +175,16 @@ tool:
...
+          - command:
+              command: "JENKINS_HOME_LINUX=\"/var/jenkins\"\nJENKINS_HOME_WINDOWS=\"\
+                C:/Users/Administrator/jenkins\"\n \n\nif uname -s | grep -i NT; then\n\
+                \    echo windows agent\n    JENKINS_JDK14_PATH=$JENKINS_HOME_WINDOWS/tools/hudson.model.JDK/openjdk-14\n\
+                \    JDK14_ZIP=\"OpenJDK14U-jdk_x64_windows_hotspot_14.0.2_12.zip\"\
+                \n    \nelse\n     echo linux agent\n     JENKINS_JDK14_PATH=$JENKINS_HOME_LINUX/tools/hudson.model.JDK/openjdk-14\n\
+                \     JDK14_ZIP=\"OpenJDK14U-jdk_x64_linux_hotspot_14.0.2_12.zip\"\
+                \n     \nfi\n\n\nmkdir -p $JENKINS_JDK14_PATH\ncd $JENKINS_JDK14_PATH\n\
+                pwd\nrm -rf ./*\ncurl -sSL https://ci.opensearch.org/ci/dbc/tools/$JDK14_ZIP\
+                \ -o jdk14.zip\nunzip -q jdk14.zip\nrm jdk14.zip\n\n"
```
gaiksaya, line 187: How will this be differentiated for windows and linux since baseJenkins is same for both?

resources/baseJenkins.yaml
```diff
@@ -175,8 +175,16 @@ tool:
...
-              url: "https://ci.opensearch.org/ci/dbc/tools/OpenJDK14U-jdk_x64_linux_hotspot_14.0.2_12.zip"
+          - command:
+              command: "JENKINS_HOME_LINUX=\"/var/jenkins\"\nJENKINS_HOME_WINDOWS=\"\
+                C:/Users/Administrator/jenkins\"\n \n\nif uname -s | grep -i NT; then\n\
+                \    echo windows agent\n    JENKINS_JDK14_PATH=$JENKINS_HOME_WINDOWS/tools/hudson.model.JDK/openjdk-14\n\
+                \    JDK14_ZIP=\"OpenJDK14U-jdk_x64_windows_hotspot_14.0.2_12.zip\"\
+                \n    \nelse\n     echo linux agent\n     JENKINS_JDK14_PATH=$JENKINS_HOME_LINUX/tools/hudson.model.JDK/openjdk-14\n\
+                \     JDK14_ZIP=\"OpenJDK14U-jdk_x64_linux_hotspot_14.0.2_12.zip\"\
+                \n     \nfi\n\n\nmkdir -p $JENKINS_JDK14_PATH\ncd $JENKINS_JDK14_PATH\n\
+                pwd\nrm -rf ./*\ncurl -sSL https://ci.opensearch.org/ci/dbc/tools/$JDK14_ZIP\
```
gaiksaya, line 186: Why are we using ci.opensearch.org? We are trying to keep this ci as generic as possible avoid any urls/artifacts related to OpenSearch

resources/baseJenkins.yaml
gaiksaya: My bad forgot that those are agents and this runs on main node.

peterzhuamazon: We use uname to differentiate between linux and windows.

prudhvigodithi (approved):

resources/baseJenkins.yaml
gaiksaya: [Offline discussion] We will be creating an issue to take this input from user rather than hardcoding all the commands. Something similar to https://github.com/opensearch-project/opensearch-ci/pull/136/files

peterzhuamazon: There are some modifications on the zip so it can install without further tweaks.
I will switch from ci to adoptium links and will open an issue to add a custom entry for tweaking global tools.

Ex: https://github.com/opensearch-project/opensearch-ci/pull/136
Thanks.

peterzhuamazon:
New follow up enhancement issue here:
* https://github.com/opensearch-project/opensearch-ci/issues/216

peterzhuamazon:
Adding more context:
* We need to add jdk14 is because we need it for LINUX/Windows gradle check, task bwcTest with 1.x version clusters. Once we do not support 1.x anymore, we can safely remove this temporary solution.
* As for why jdk14 cannot be automatically install on jenkins, see here: https://github.com/opensearch-project/opensearch-build/issues/2604.
* We move the artifacts on jenkins is to make sure it is consistent between linux and windows version of the jdk, and strip the leading subfolder so jenkins can natually unzip it in the same directly.
* Since jdk14 is deprecated now we also need to preserve the binaries for the time being.

Thanks.

Issue #217, opened 2022-11-02
Bug: Fix cloudwatch alarm `MainNodeTooManyJenkinsProcessesFound`


prudhvigodithi:
### Describe the bug

The existing alarm `MainNodeTooManyJenkinsProcessesFound` sometimes shows false positives, the configured [procstat](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch-Agent-procstat-process-metrics.html) alarm assumes there are more than one jenkins process.

Solution: 
Change the alarm in way instead of monitoring more than one process, monitor the main jenkins process and alert if the procstat java process `procstat_lookup_pid_count` jenkins.war is down.
This way to trigger the alarm when jenkins is down makes more sense.

### To reproduce

N/A

gaiksaya:
closed via https://github.com/opensearch-project/opensearch-ci/pull/227

PR #218, opened 2022-11-02
Bug: Fix cloudwatch alarm `MainNodeTooManyJenkinsProcessesFound`

prudhvigodithi:
Signed-off-by: Prudhvi Godithi <pgodithi@amazon.com>

### Description
Fix cloudwatch alarm `MainNodeTooManyJenkinsProcessesFound` to `MainNodeJenkinsProcessNotFound`

### Issues Resolved
https://github.com/opensearch-project/opensearch-ci/issues/217

### DIFF
--- a/lib/monitoring/ci-alarms.ts
+++ b/lib/monitoring/ci-alarms.ts
@@ -44,12 +44,13 @@ export class JenkinsMonitoring {
       treatMissingData: TreatMissingData.BREACHING,
     }));
 
-    this.alarms.push(new Alarm(stack, 'MainNodeTooManyJenkinsProcessesFound', {
-      alarmDescription: 'Only one jenkins process should run at any given time on the main node, there might be a cloudwatch configuration issue',
-      metric: mainNode.ec2InstanceMetrics.foundJenkinsProcessCount.with({ statistic: 'max' }),
-      evaluationPeriods: 3,
+    this.alarms.push(new Alarm(stack, 'MainNodeJenkinsProcessNotFound', {
+      alarmDescription: 'Jenkins process is not running',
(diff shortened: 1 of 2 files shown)


peterzhuamazon (approved):

lib/monitoring/ci-alarms.ts
```diff
@@ -44,12 +44,13 @@ export class JenkinsMonitoring {
       treatMissingData: TreatMissingData.BREACHING,
     }));
 
-    this.alarms.push(new Alarm(stack, 'MainNodeTooManyJenkinsProcessesFound', {
-      alarmDescription: 'Only one jenkins process should run at any given time on the main node, there might be a cloudwatch configuration issue',
-      metric: mainNode.ec2InstanceMetrics.foundJenkinsProcessCount.with({ statistic: 'max' }),
-      evaluationPeriods: 3,
+    this.alarms.push(new Alarm(stack, 'MainNodeJenkinsProcessNotFound', {
+      alarmDescription: 'Jenkins process is not running',
+      metric: mainNode.ec2InstanceMetrics.foundJenkinsProcessCount.with({ statistic: 'avg' }),
```
rishabh6788, line 49: is `avg` the right stat? Should we change it to `sum` to make sure at any point of time total number of process should not be less than sum of all the process count at that data point, which should always be 1

prudhvigodithi: @rishabh6788 thanks I believe sum might not work here, sometimes the value could be 0.89, the average to check to for 1 is what is right, as we should only check 1 or 0 up or down and trigger an alarm.

prudhvigodithi: I have also tested the alarm , to stop and start the jenkins container works as expected

rishabh6788 (approved):

PR #252, opened 2023-02-16
Add grub in apt-mark so Ubuntu2004 wont update

### LINKED ISSUE #251: Ubuntu2004 AMI keeps having issues when upgrading grub2 bootloader
Ubuntu2004 constantly update the grub2 bootloader pkg which requires manual interations.
With noninteractive flag it will fail the installation.

Therefore we need to disable grub2 upgrade during non-interactive shell script runs. Thanks.

peterzhuamazon:
### Description
Add grub in apt-mark so Ubuntu2004 wont update

### Issues Resolved
#251

### DIFF
--- a/lib/compute/agent-nodes.ts
+++ b/lib/compute/agent-nodes.ts
@@ -112,8 +112,7 @@ export class AgentNodes {
       numExecutors: 1,
       amiId: 'ami-042cbc70322ec7fd5',
       initScript: 'sudo apt-mark hold docker docker.io openssh-server gh grub-efi* && docker ps &&'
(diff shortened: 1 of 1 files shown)


jordarlu (approved):

gaiksaya:
Is there a way to run it with non-interative shell?
Some quick search got me this https://askubuntu.com/questions/531364/non-interactive-install-of-grub2-when-installing-from-scratch

peterzhuamazon:
> Is there a way to run it with non-interative shell? Some quick search got me this https://askubuntu.com/questions/531364/non-interactive-install-of-grub2-when-installing-from-scratch

We already use the non-interactive shell and most of the time grub without interactive shell will goes to failure anyway. This has been tested and we do not need to upgrade grub2 that often.

Thanks.

PR #262, opened 2023-03-29
Make Jenkins use private subnet for all nodes

### LINKED ISSUE #258: Create a CI workflow that creates new AMIs using packer

Currently the AMI's used as agent nodes using a specific bas image that may go out of date or needs updates as new kernel updates come in. 
This happens as often as per quarter. Even though we run yum update, apt updates, etc we still need to reboot the EC2 to apply those updates which does not fit jenkins' agent nodes' lifecycle management. If a SSH connection is lost (when we reboot) a new agent will be brought up.

In order to apply regular updates to the base AMI image we need to build a new AMI. 
Using packer it is a pretty straight forward process. https://github.com/opensearch-project/opensearch-ci/tree/main/packer

Below are 2 possible approaches:
1. Use GHA that will create new AMI's and create a pull request to update the same in this repository
2. Use jenkins workflow that will do the same.

Please keep in mind that this needs to be a blue green deployment and that's why old AMI's need to be deprecated (made private) only after confirming new AMI's are working fine. This can be a will be a manual process to start with but can be automated via GHA too if we maintain a list somewhere. 

### Describe alternatives you've considered

Do the entire process manually. However building AMI even using packer takes more than half a day for the number of AMI's we have.

### Additional context

_No response_

peterzhuamazon:
### Description
Make Jenkins use private subnet for all nodes

### Issues Resolved
#258

### DIFF
--- a/lib/ci-stack.ts
+++ b/lib/ci-stack.ts
@@ -175,7 +175,7 @@ export class CIStack extends Stack {
       failOnCloudInitError: props?.ignoreResourcesFailures,
       adminUsers: props?.adminUsers,
       agentNodeSecurityGroup: this.securityGroups.agentNodeSG.securityGroupId,
-      subnetId: vpc.publicSubnets[0].subnetId,
+      subnetId: vpc.privateSubnets[0].subnetId,
     }, this.agentNodes, macAgentParameter.toString(), props?.agentAssumeRole);
 
     const externalLoadBalancer = new JenkinsExternalLoadBalancer(this, {


peterzhuamazon:
I have verified that the Jenkins can run no issues with all agent nodes in private subnet.
This change would prevent any further external connection to nodes in jenkins.
Thanks.

zelinh (approved):

gaiksaya (approved):

Issue #346, opened 2023-09-08
Jenkins main node has very high cpu and mem usages


peterzhuamazon:
Jenkins spikes on cpu and mem usages for a while.
This is the result of a lot of benchmark heavyweight test + distribution build + integTest.

We want to increase the cpu and mem base by switching to a different instance type.

```
CONTAINER ID   NAME      CPU %     MEM USAGE / LIMIT   MEM %     NET I/O          BLOCK I/O     PIDS
3b3f79448b93   jenkins   746.52%   19.95GiB / 20GiB    99.77%    37.4GB / 163GB   40.9MB / 0B   686
```

We also want to increase the stack size of Jenkins java process as it is currently utilizing the default 512kb stack size.
https://issues.jenkins.io/browse/JENKINS-52966

Thanks.

peterzhuamazon:
Need to finetune later with this:
https://www.jenkins.io/blog/2016/11/21/gc-tuning/

peterzhuamazon:
Error related to this one here:
```

Invalid maximum heap size: -Xmx60g -Xss4m
Error: Could not create the Java Virtual Machine.
Error: A fatal exception has occurred. Program will exit.

```

Related: 
* https://github.com/jenkinsci/docker/issues/213
* https://github.com/jenkinsci/docker/issues/202

peterzhuamazon:
Will remove quotes.

peterzhuamazon:
We still need to address the gc as Jenkins just keep growing on the mem usages even with higher specs on EC2 instance type:
```
CONTAINER ID   NAME      CPU %     MEM USAGE / LIMIT   MEM %     NET I/O         BLOCK I/O     PIDS
6ede06ab221a   jenkins   25.67%    39.28GiB / 64GiB    61.38%    102GB / 482GB   0B / 30.2kB   388
```

Probably switch to the g1gc setups.

peterzhuamazon:
We can look into the monitoring plugin:
https://plugins.jenkins.io/monitoring/

It has manual click on webpage as well as scheduled gc functionalities in the plugin here.

peterzhuamazon:
The plugin is really useful right here:

Before:
```
13.09GiB / 64GiB    20.45%
```
After:
```
3.781GiB / 64GiB 
```
[...]

PR #353, opened 2023-10-02
Improve jenkins performance

prudhvigodithi:
### Description
Improve jenkins performance, this should address the jenkins 504 timeout errors due to `OutOfMemoryError`

### Issues Resolved
Part of https://github.com/opensearch-project/opensearch-ci/issues/346

**Analysis-1**

Started to Debug with created a 360 java heap dump using an open source [yc tool](https://github.com/ycrash/yc-data-script).

Extracted the yc-2023-09-29T04-32-09.zip and used jhat cli debug the heap dump, noted error as: Exception in thread "main" java.lang.OutOfMemoryError: Java heap space.

Also analyzed all the dump files part of the zip and found out multiple threads in `BLOCKED` state each honoring the setting `-Xss4m`.

Apart form having the threads in `BLOCKED` state. noted multiple `I/O` errors with threads interrupted state.

**Took some measures to fix this.**

* Removed the `-Xss` setting to avoid large overloaded thread stack size.
* Increased the docker memory and application Xmx limits (applied manually to test and observe if this would resolve the frequent OutOfMemoryError error).
* Deleted all agent nodes to clean up the I/O errors and threads in interrupted state.
* Restarted docker (jenkins) container.

**Analysis-2**

Coming from jenkins GC performance tuning, added the following to the jenkins Jvm.
```
- JENKINS_JAVA_OPTS=-Xms33g -Xmx65g -Dhudson.model.ParametersAction.keepUndefinedParameters=true -XX:+UseG1GC -XX:+ExplicitGCInvokesConcurrent -XX:+ParallelRefProcEnabled -XX:+UseStringDeduplication -XX:+UnlockExperimentalVMOptions -XX:G1NewSizePercent=20 -XX:+UnlockDiagnosticVMOptions -XX:G1SummarizeRSetStatsPeriod=1
```

`-XX:+UseG1GC`: Changes to Garbage First Garbage Collector strategy, G1GC is designed to be a low-latency garbage collector. It divides the heap into regions and performs garbage collection on specific regions.
`-XX:+ExplicitGCInvokesConcurrent`: This flag allows explicit garbage collection calls (System.gc()) to run concurrently with the application threads.
`-XX:+ParallelRefProcEnabled`: This flag enables parallel reference processing which enables us to use the multiple cores more effectively.
`-XX:+UseStringDeduplication`: This option enables string deduplication in G1GC. It means that duplicate strings in the heap will be deduplicated, helping to save memory by sharing the same underlying string data.
`-XX:+UnlockExperimentalVMOptions`: This flag allows the use of experimental JVM options.
`-XX:+UnlockDiagnosticVMOptions`: This flag unlocks diagnostic VM options, allowing the use of additional diagnostic options for monitoring and troubleshooting the JVM.
`-XX:G1SummarizeRSetStatsPeriod=1`: This option sets the frequency at which G1GC summarizes.

Along with above Jvm options, Manually identified the running threads using the following script and noticed there are some running threads in errored state.

```
import jenkins.model.Jenkins
import hudson.model.Queue

def jenkins = Jenkins.getInstance()

// Get the Jenkins queue
def queue = jenkins.getQueue()

// Get all items in the queue
def items = queue.getItems()

// Loop through each item in the queue
items.each { item ->
    println "Queue Item: ${item.task.name}, In Queue Since: ${item.inQueueSince}"

    // Get the buildable task associated with the queue item
    def buildable = item.getParams()

    if (buildable) {
        println "  Buildable Task: ${buildable.displayName}"
    }
    
    println ""
}

// Get all active threads
def threads = Thread.getAllStackTraces().keySet()

println "\nRunning Threads:"
threads.each { thread ->
    println "  Thread Name: ${thread.getName()}"
    println "  Thread ID: ${thread.getId()}"
    println "  Thread State: ${thread.getState()}"
    println ""
}
```
Now Executed `interrupt()` method on each thread and manually killed it. Then after restarting the jenkins the memory usage is stable and did not notice any 504 errors.

NOTE: Used `htop` linux tool to monitor the memory in real time.

### DIFF
--- a/resources/docker-compose.yml
+++ b/resources/docker-compose.yml
@@ -11,7 +11,7 @@ services:
       - 50000:50000
     container_name: jenkins
     environment:
-      - JENKINS_JAVA_OPTS=-Xmx60g -Xss4m
+      - JENKINS_JAVA_OPTS=-Xms32g -Xmx64g -Dhudson.model.ParametersAction.keepUndefinedParameters=true -XX:+UseG1GC -XX:+ExplicitGCInvokesConcurrent -XX:+ParallelRefProcEnabled -XX:+UseStringDeduplication -XX:+UnlockExperimentalVMOptions -XX:G1NewSizePercent=20 -XX:+UnlockDiagnosticVMOptions -XX:G1SummarizeRSetStatsPeriod=1
       - CASC_RELOAD_TOKEN=reloadPasswordHere
     volumes:
       - /var/lib/jenkins:/var/jenkins_home
@@ -19,10 +19,10 @@ services:
       resources:
         limits:
           cpus: '32'
-          memory: '64g'
+          memory: '65g'
         reservations:
           cpus: '32'
-          memory: '64g'
+          memory: '65g'
     logging:
       driver: awslogs
       options:


prudhvigodithi:
> Should we take care of upgrading the main node instance size? https://github.com/opensearch-project/opensearch-ci/blob/main/lib/compute/jenkins-main-node.ts#L115

We already have c5.9xlarge (68gb memory and 36 CPU), this should be sufficient with the limits I have added.

gaiksaya (approved):
LGTM! Let's try this out.

resources/docker-compose.yml
```diff
@@ -11,18 +11,18 @@ services:
...
         limits:
-          cpus: '32'
-          memory: '64g'
+          cpus: '34'
+          memory: '67g'
         reservations:
-          cpus: '32'
-          memory: '64g'
+          cpus: '34'
+          memory: '67g'
```
peterzhuamazon, line 25: Thanks @prudhvigodithi .

Can we keep the cpu and xms to 32/32g?
And xmx 64g?

So the host has a few more resources even if docker is throttling the instance. Thanks.

prudhvigodithi: Updated the PR to have some bandwidth for the server process. Thanks

Issue #321, opened 2023-07-07
Investigate the Spike in Flaky test failures as a function of the gradle check configuration and Jenkins Runniner instance sizing


nknize:
### Is your feature request related to a problem? Please describe

Coming out of this [public slack discussion](https://opensearch.slack.com/archives/C04UTNM338A/p1688765685170159) I'd like to explore a possible spike in flaky test failures during `gradlew check` on PRs in the [OpenSearch core repository](https://github.com/opensearch-project/opensearch) during regular business hours. 

The concrete test failures we're noticing are similar to:

```
Caused by: java.net.ConnectException: Connection refused
	at sun.nio.ch.Net.pollConnect(Native Method) ~[?:?]
	at sun.nio.ch.Net.pollConnectNow(Net.java:672) ~[?:?]
	at sun.nio.ch.SocketChannelImpl.finishConnect(SocketChannelImpl.java:946) ~[?:?]
	at org.opensearch.nio.SocketChannelContext.connect(SocketChannelContext.java:157) ~[opensearch-nio-2.9.0-SNAPSHOT.jar:2.9.0-SNAPSHOT]
```

As can be seen in this [one instance](https://build.ci.opensearch.org/job/gradle-check/19457/testReport/junit/org.opensearch.search/SearchWeightedRoutingIT/testStrictWeightedRoutingWithCustomString/). This seems mostly related to socket issues in the runner and seems to occur on "aggressive" Integration Tests (e.g., those using `Scope.Test` level, which fires up a new cluster for each test method). 

With jenkins having its own Runner for each invocation I wouldn't expect the high level of activity (e.g., multiple PRs throughout the day) to contribute, so maybe this is more related to the test intensity, `--parallel` gradle invocation, and size of the Runner instance? 

### Describe the solution you'd like

As a parallel effort to trying to lean out the intense integration tests in the core repo, I'd like for us to see if we can root cause these time outs as a function of instance resources (e.g., CPU, Memory) and the test configuration (e.g., number of concurrent integration tests, number of sockets).

It may be that we just aren't closing the sockets in the core IntegrationTest class? (we can explore that separately). 

### Describe alternatives you've considered

* Check the core Integration Test harness is properly closing sockets
* Check the socket pool configuration in the core test framework.
* ... other core improvements not explicitly mentioned here.

### Additional context

Thank you!

peterzhuamazon:
[...]

PR #355, opened 2023-10-03
Remove old Windows runner and replace Windows Gradle Check Runner AMI ID with M58xlarge

peterzhuamazon:
### Description
Remove old Windows runner and replace Windows Gradle Check Runner AMI ID with M58xlarge

### Issues Resolved
https://github.com/opensearch-project/opensearch-build/issues/3816

### DIFF
--- a/lib/compute/agent-nodes.ts
+++ b/lib/compute/agent-nodes.ts
@@ -32,8 +32,6 @@ export class AgentNodes {
 
   readonly MACOS12_X64_MULTI_HOST: AgentNodeProps;
 
-  readonly WINDOWS2019_X64: AgentNodeProps;
-
   readonly WINDOWS2019_X64_DOCKER_HOST: AgentNodeProps;
 
   readonly WINDOWS2019_X64_DOCKER_BUILDER: AgentNodeProps;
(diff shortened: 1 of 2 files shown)


lib/compute/agent-nodes.ts
```diff
@@ -184,19 +182,6 @@ export class AgentNodes {
       initScript: 'echo',
       remoteFs: '/var/jenkins',
     };
-    this.WINDOWS2019_X64 = {
```
prudhvigodithi, line 187: We no longer use this @peterzhuamazon ?

prudhvigodithi (approved):

lib/compute/agent-nodes.ts
```diff
@@ -234,13 +219,13 @@ export class AgentNodes {
     this.WINDOWS2019_X64_GRADLE_CHECK = {
       agentType: 'windows',
       customDeviceMapping: '/dev/sda1=:300:true:::encrypted',
-      workerLabelString: 'Jenkins-Agent-Windows2019-X64-C524xlarge-Single-Host',
-      instanceType: 'C524xlarge',
+      workerLabelString: 'Jenkins-Agent-Windows2019-X64-M58xlarge-Single-Host',
+      instanceType: 'M58xlarge',
```
prudhvigodithi, line 223: Whey are we changing the instance type, is C5 type not stable ?

lib/compute/agent-nodes.ts
peterzhuamazon: Nope, we have since changed to windows docker in 2.10.0.

lib/compute/agent-nodes.ts
peterzhuamazon: Yes, c524large is 1:2 cpu:mem while m58xlarge is 1:4 cpu:mem, which linux has already replace with.
Runs better, more stable, and less resource.

* #321

PR #359, opened 2023-10-05
Update Windows Gradle Check Runner to use 126GB WINRM Mem on M58xlarge

peterzhuamazon:
### Description
Update Windows Gradle Check Runner to use 126GB WINRM Mem on M58xlarge

### Issues Resolved
https://github.com/opensearch-project/opensearch-build/issues/3816

### DIFF
--- a/packer/jenkins-agent-win2019-x64-gradle-check.json
+++ b/packer/jenkins-agent-win2019-x64-gradle-check.json
@@ -68,7 +68,7 @@
     {
       "type":"powershell",
       "inline": [
-        "C:\\Users\\Administrator\\jenkins\\winrm_max_memory.ps1 124"
+        "C:\\Users\\Administrator\\jenkins\\winrm_max_memory.ps1 126"
       ]
     },
     {
--- a/packer/jenkins-agent-win2019-x64-gradle-check.json
+++ b/packer/jenkins-agent-win2019-x64-gradle-check.json
@@ -87,7 +87,12 @@
     {
       "type":"powershell",
       "inline": [
-        "C:\\ProgramData\\Amazon\\EC2-Windows\\Launch\\Scripts\\InitializeInstance.ps1 -Schedule",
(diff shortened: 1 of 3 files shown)


packer/scripts/windows/winrm_max_memory.ps1
```diff
@@ -5,6 +5,9 @@
 # this file be licensed under the Apache-2.0 license or a
 # compatible open source license.
 
+echo "The max amount of the winrm memory must be 2GB smaller than max host memory, no less"
```
zelinh, line 8: A little confused, so for example, the winrm memory has to be set more than 126 if the max is 128? or it has to be exact 126.

peterzhuamazon: I think it is minimum to be 126G.

packer/jenkins-agent-win2019-x64-gradle-check.json
```diff
@@ -87,7 +87,12 @@
     {
       "type":"powershell",
       "inline": [
-        "C:\\ProgramData\\Amazon\\EC2-Windows\\Launch\\Scripts\\InitializeInstance.ps1 -Schedule",
+        "C:\\ProgramData\\Amazon\\EC2-Windows\\Launch\\Scripts\\InitializeInstance.ps1 -Schedule"
```
zelinh, line 90: Note: so we split running these two scripts to make sure we are running one by one instead of together to avoid race condition?

peterzhuamazon: Yes, seems like right now the second script run success will overshadow the failure of the first script at some point.
So just separate them out so we will have a consistent outcome.

Thanks.

zelinh (approved):

peterzhuamazon:
https://build.ci.opensearch.org/job/packer-build/77/console

PR #378, opened 2023-12-12
Redirect cf-url-rewriter to fetch respective index.json file

Divyaasm:
### Description
Redirect cf-url-rewriter to fetch latest buildnumber from the index.json file corresponding to the given distribution type in the url

### Issues Resolved
https://github.com/opensearch-project/opensearch-build/issues/4166


rishabh6788 (approved):

rishabh6788:
Question: Have you verified downloads with existing/older artifacts?

Divyaasm:
> Question: Have you verified downloads with existing/older artifacts?

Yes Rishabh, this change is compatible new folder structure `index`. uploadIndexFile library will now store the latest build number of a distribution in it's respective sub-folder inside index folder

tianleh:
My original [PR](https://github.com/opensearch-project/opensearch-build/pull/1569/files#diff-6e706a04cf9f6f687023d96f14a96c4ee0b2f4369646875978f2e903c13d71e6) when implementing the `latest` feature came with unit tests. Looks like the test files were not moved over from `build` repo to this `ci` repo.

Could you make the move and update the test files?

@Divyaasm

tianleh:
> Yes Rishabh, this change is compatible new folder structure `index`. uploadIndexFile library will now store the latest build number of a distribution in it's respective sub-folder inside index folder

Have a similar question as @rishabh6788 What about the `latest` url for an old version, e.g 1.2.0 which may not have a recent build and thus still have an old structure?

Divyaasm:
> > Yes Rishabh, this change is compatible new folder structure `index`. uploadIndexFile library will now store the latest build number of a distribution in it's respective sub-folder inside index folder
> 
> Have a similar question as @rishabh6788 What about the `latest` url for an old version, e.g 1.2.0 which may not have a recent build and thus still have an old structure?

The index file folder is updated in the corresponding build jenkins workflow when triggered. Since we do not touch the older versions they remain the same. And coming to the latest running jobs the path in s3 bucket is updated and retrieved.

resources/cf-url-rewriter/cf-url-rewriter.ts
```diff
@@ -66,3 +73,12 @@ function errorResponse() {
     statusDescription: 'Not found',
   };
 }
+
+function getIndexFilePath(versionNumber: string, urisplit: string[]): string {
+    if (versionNumber <= "2111") {
```
tianleh, line 78: Based on the release history https://github.com/opensearch-project/opensearch-build/releases , I see that we are still releasing `1.3.x` If we have a `1.3.15` release, looks like the logic here will still use the old index.json path.

resources/cf-url-rewriter/cf-url-rewriter.ts
```diff
@@ -19,9 +19,16 @@ export async function handler(event: CloudFrontRequestEvent, context: Context, c
 
   if (request.uri.includes("/latest/")) {
 
-    const indexUri = request.uri.replace(/\/latest\/.*/, '/index.json');
+    const urisplit = request.uri.split("/latest");
+    const index = urisplit[0].lastIndexOf("/");
+    const versionNumber = urisplit[0].substring(index + 1).replace(/\./g, '');
```
tianleh, line 24: nonblocking: all these variables are used for `getIndexFilePath` and not in the main function. Better to move them into the sub function getIndexFilePath

resources/cf-url-rewriter/cf-url-rewriter.ts
Divyaasm: Just updated the requested changes. Can you please review again @tianleh. Thanks

resources/cf-url-rewriter/cf-url-rewriter.ts
```diff
@@ -66,3 +71,20 @@ function errorResponse() {
...
 }
+
+function getIndexFilePath(urisplit: string[]): string {
+    const version = urisplit[0].substring(urisplit[0].lastIndexOf("/") + 1);
+    const [major, minor, patch] = version.split('.').map(Number);
+
+    if (
+        (major == 1 && minor == 3 && patch >= 15) ||
+        (major == 2 && minor >= 12) ||
+        (major >= 3)
```
tianleh, line 82: It is indeed a tricky problem. The latest patch release for `2.11` is `2.11.1` I think it is theoretically possible to have a newer patch release, say `2.11.2`. Then this logic will fail. The same thing will happen to all existing 2.x releases < 2.12

So we have several options here.

Option 1: extend your current comparison logic to cover all 2.x versions. This will be a long list.

Option 2: back fill S3 buckets for all released versions with the new structure. Then you can remove the comparison. This can simplify the code but require tedious manual effort.

Option 3: always run httpsGet against the new url. If returned empty, then fall back to the old url. Then you do not need to backfill S3 buckets and also do not need to maintain a long comparison list. 

I would prefer Option 3 for simplicity.

resources/cf-url-rewriter/cf-url-rewriter.ts
tianleh: Commented https://github.com/opensearch-project/opensearch-ci/pull/378#discussion_r1442314387

resources/cf-url-rewriter/cf-url-rewriter.ts
gaiksaya: If you want to get the semver try using https://www.npmjs.com/package/semver library.

resources/cf-url-rewriter/cf-url-rewriter.ts
```diff
@@ -19,21 +19,31 @@ export async function handler(event: CloudFrontRequestEvent, context: Context, c
...
   if (request.uri.includes("/latest/")) {
 
-    const indexUri = request.uri.replace(/\/latest\/.*/, '/index.json');
+    const latestPath = request.uri.split("/latest")[1].split("/").slice(0, 4).join("/");
 
-    try {
-      const data: any = await httpsGet('https://' + request.headers.host[0].value + indexUri);
+    const latestIndexUri = request.uri.replace(/\/latest\/.*/, '/index' + latestPath + '/index.json');
+
+    try{
```
tianleh, line 26: missing indentation

resources/cf-url-rewriter/cf-url-rewriter.ts
```diff
@@ -19,21 +19,31 @@ export async function handler(event: CloudFrontRequestEvent, context: Context, c
...
-      } else {
-        callback(null, errorResponse());
       }
     } catch (e) {
-      console.log(e);
-      callback(null, errorResponse());
-    }
+      try{
+      const oldIndexUri = request.uri.replace(/\/latest\/.*/, '/index.json');
+      const indexData: any = await httpsGet('https://' + request.headers.host[0].value + oldIndexUri);
```
tianleh, line 35: missing indentation. Please format the code

resources/cf-url-rewriter/cf-url-rewriter.ts
```diff
@@ -19,21 +19,31 @@ export async function handler(event: CloudFrontRequestEvent, context: Context, c
...
+          callback(null, redirectResponse(request, indexData.latest));
+        } else {
+          callback(null, errorResponse());
+        }
+      } catch (e) {
+        console.log(e);
+        callback(null, errorResponse());
+      }
+    }
   } else {
```
tianleh, line 47: You can simplify the code by extracting common logic to a sub function. Below is the skeleton.

```

function getLatest(url) {

  try {
    const data: any = await httpsGet(url);

    if (data && data.latest) {
      return data.latest;
    }

    return "";
  } catch (e) {
    console.log(e);
    return "";
  }

}

function handler(() {

  ...
  const latestFromNewUrl = getLatest(newUrl);

  if (latestFromNewUrl) {
    callback(null, redirectResponse(request, latest));
  } else {
    const latestFromOldUrl = getLatest(oldUrl);

    if (latestFromOldUrl) {
      callback(null, redirectResponse(request, latestFromOldUrl));
    } else {
      callback(null, errorResponse());
    }

  }

  ...
}
```

tianleh (approved):

rishabh6788:
@Divyaasm Do we need this now?

Divyaasm:
> @Divyaasm Do we need this now?

This will be used for deploying the the cf-url changes using opensearch-ci repo. I need to make the PR similar to the internal changes. We can have the test cases too here.

rishabh6788:
LGTM!

Issue #386, opened 2024-02-02
[Bug]: Jenkins critical CVE caused by built-in CLI vulnerabilities - mitigation method


jordarlu:
### Describe the bug

[Jenkins Security Advisory 2024-01-24](https://www.jenkins.io/security/advisory/2024-01-24/#jenkins-security-advisory-2024-01-24) described a critical CVE that applied to all Jenkins controller versions 2.441 and earlier.
The current CI Jenkins controller version is 2.387.1, and we can take the [suggested steps from Jenkins to mitigate the issue ](https://www.jenkins.io/doc/book/managing/groovy-hook-scripts/)for the time being. 

### To reproduce

This CVE is applicable to all Jenkins controller versions 2.441 and earlier.

### Expected behavior

We can take the following steps suggested by Jenkins to mitigate issue: 

> Fix Description:
> Jenkins 2.442, LTS 2.426.3 disables the command parser feature that replaces an @ character followed by a file path in an argument with the file’s contents for CLI commands.
> 
> In case of problems with this fix, disable this change by setting the Java system property hudson.cli.CLICommand.allowAtSyntax to true. Doing this is strongly discouraged on any network accessible by users who are not Jenkins administrators.
> 
> Workaround:
> Disabling access to the CLI is expected to prevent exploitation completely. Doing so is strongly recommended to administrators unable to immediately update to Jenkins 2.442, LTS 2.426.3. Applying this workaround does not require a Jenkins restart. For instructions, see the documentation for this workaround.   ( https://github.com/jenkinsci-cert/SECURITY-3314-3315/ )

also, refer to here (https://www.jenkins.io/doc/book/managing/groovy-hook-scripts/) to create init.groovy.d folder and palce the disable-cli.groovy

jordarlu:
close this issue as the PR has been merged.

PR #388, opened 2024-02-06
Jenkins CLI CVE mitigation

jordarlu:
### Description
Jenkins CLI CVE https://www.jenkins.io/security/advisory/2024-01-24/ mitigation

### Issues Resolved
https://github.com/opensearch-project/opensearch-ci/issues/386

### DIFF
--- a/lib/compute/jenkins-main-node.ts
+++ b/lib/compute/jenkins-main-node.ts
@@ -268,7 +268,7 @@ export class JenkinsMainNode {
       // Jenkins CVE https://www.jenkins.io/security/advisory/2024-01-24/ mitigation 
       InitCommand.shellCommand('mkdir -p /var/lib/jenkins/init.groovy.d'),
       // eslint-disable-next-line max-len
-      InitCommand.shellCommand('sudo wget -P /var/lib/jenkins/init.groovy.d https://raw.githubusercontent.com/jenkinsci-cert/SECURITY-3314-3315/main/disable-cli.groovy'),
+      InitCommand.shellCommand('sudo curl -SL https://raw.githubusercontent.com/jenkinsci-cert/SECURITY-3314-3315/main/disable-cli.groovy -o /var/lib/jenkins/init.groovy.d/disable-cli.groovy'),
 
       // Configuration to proxy jenkins on :8080 -> :80
       InitFile.fromString('/etc/httpd/conf.d/jenkins.conf',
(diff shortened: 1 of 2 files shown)


jordarlu:
Validated at personal public AWS account, the Jenkins CLI page resulted in 404 Not Found error as expected.

lib/compute/jenkins-main-node.ts
```diff
@@ -265,6 +265,11 @@ export class JenkinsMainNode {
       + ' instance_id=`curl -f -H "X-aws-ec2-metadata-token: $TOKEN" http://169.254.169.254/latest/meta-data/instance-id` && echo $ami_id &&'
       + ` aws ec2 --region ${stackRegion} modify-instance-metadata-options --instance-id $instance_id --http-put-response-hop-limit 2`),
 
+      // Jenkins CVE https://www.jenkins.io/security/advisory/2024-01-24/ mitigation 
+      InitCommand.shellCommand('mkdir -p /var/lib/jenkins/init.groovy.d'),
+      // eslint-disable-next-line max-len
+      InitCommand.shellCommand('sudo wget -P /var/lib/jenkins/init.groovy.d https://raw.githubusercontent.com/jenkinsci-cert/SECURITY-3314-3315/main/disable-cli.groovy'),
```
peterzhuamazon, line 271: Suggest using curl instead of wget as wget is not always installed by default. Also on curl you can try to use `curl -SL <url> -o <path/to/file>`.

Tho not hard block on this.

peterzhuamazon (approved):

lib/compute/jenkins-main-node.ts
```diff
@@ -268,7 +268,7 @@ export class JenkinsMainNode {
       // Jenkins CVE https://www.jenkins.io/security/advisory/2024-01-24/ mitigation 
       InitCommand.shellCommand('mkdir -p /var/lib/jenkins/init.groovy.d'),
       // eslint-disable-next-line max-len
-      InitCommand.shellCommand('sudo wget -P /var/lib/jenkins/init.groovy.d https://raw.githubusercontent.com/jenkinsci-cert/SECURITY-3314-3315/main/disable-cli.groovy'),
+      InitCommand.shellCommand('sudo curl -SL https://raw.githubusercontent.com/jenkinsci-cert/SECURITY-3314-3315/main/disable-cli.groovy -o /var/lib/jenkins/init.groovy.d/disable-cli.groovy'),
```
peterzhuamazon, line 271: Another issue is we should use a commit to lock this file, instead of branch `main` as that would be a moving target.

lib/compute/jenkins-main-node.ts
jordarlu: it is neat! ,,thanks,, I changed to curl -SL -o, will merge it!!

lib/compute/jenkins-main-node.ts
jordarlu: thanks ,, makes sense, and I learned that by doing this way, we increase stability and reproducibility.. thanks for the guidence!

PR #390, opened 2024-02-19
Replace cdk-ec2-key-pair with CfnKeyPair construct to generate ec2 key pair

rishabh6788:
### Description
Replace cdk-ec2-key-pair with CfnKeyPair construct to generate ec2 key pair. 
`aws-cdk-lib` provides `CfnKeyPair` construct to generate ec2 key-pairs. The private key is saved to AWS Systems Manager Parameter Store, using a parameter with the following name: /ec2/keypair/{key_pair_id}. See https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_ec2.CfnKeyPair.html for more details. 

### Issues Resolved

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,46 +53,48 @@ OpenSearch Continuous Integration is an open source CI system for OpenSearch and
 
 ### Dev Deployment 
 1. Setup your local machine to credentials to deploy to the AWS Account
-1. Deploy the bootstrap stack by running the following command that sets up required resources to create the stacks. [More info](https://docs.aws.amazon.com/cdk/latest/guide/bootstrapping.html)
+2. Deploy the bootstrap stack by running the following command that sets up required resources to create the stacks. [More info](https://docs.aws.amazon.com/cdk/latest/guide/bootstrapping.html)
    
    `npm run cdk bootstrap -- -c useSsl=false -c runWithOidc=false -c serverAccessType=ipv4 -c restrictServerAccessTo=10.10.10.10/32`
    
-1. Deploy the ci-config-stack using the following (takes ~1 minute to deploy) - 
+3. Deploy the ci-config-stack using the following (takes ~1 minute to deploy) - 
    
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -53,46 +53,48 @@ OpenSearch Continuous Integration is an open source CI system for OpenSearch and
 
 ### Dev Deployment 
 1. Setup your local machine to credentials to deploy to the AWS Account
-1. Deploy the bootstrap stack by running the following command that sets up required resources to create the stacks. [More info](https://docs.aws.amazon.com/cdk/latest/guide/bootstrapping.html)
+2. Deploy the bootstrap stack by running the following command that sets up required resources to create the stacks. [More info](https://docs.aws.amazon.com/cdk/latest/guide/bootstrapping.html)
```
peterzhuamazon, line 56: You can keep these 1/1/1 as they will render as 1/2/3 😄

peterzhuamazon (approved):

lib/compute/agent-node-config.ts
```diff
@@ -48,12 +49,24 @@ export class AgentNodeConfig {
...
     });
-    Tags.of(key)
-      .add('jenkins:credentials:type', 'sshUserPrivateKey');
+
+    // Create a secret for storing the actaul value of key-pair so that jenkins secret-manager plugin is able to fetch it
+    const keyPairSecret = new Secret(stack, 'AgentNodeKeyPair', {
+      secretName: 'jenkins-agent-node-key-pair',
+      description: 'ssh key for jenkins main node to connect to agent nodes',
+    });
+    Tags.of(keyPairSecret).add('jenkins:credentials:type', 'sshUserPrivateKey');
```
gaiksaya, line 67: So we need this tag for both secret as well as actual ec2 key pair?

peterzhuamazon: We have to add this tag so that jenkins can read this and use it as ssh key, isnt it?

rishabh6788: This tag is mandatory for the secret but not for the actual key-pair. 
Don't see any issue with adding the same tag to key-pair as well.

PR #404, opened 2024-03-20
Add WAF rules to protect Jenkins endpoint

gaiksaya:
### Description
Adding one more layer of security by adding WAF rules and associating with the load balancer. This prevents DDoS attack when endpoint is public.

### Issues Resolved

### DIFF
--- a/lib/ci-stack.ts
+++ b/lib/ci-stack.ts
@@ -25,6 +25,7 @@ import { RunAdditionalCommands } from './compute/run-additional-commands';
 import { JenkinsMonitoring } from './monitoring/ci-alarms';
 import { JenkinsExternalLoadBalancer } from './network/ci-external-load-balancer';
 import { JenkinsSecurityGroups } from './security/ci-security-groups';
+import { JenkinsWAF } from './security/waf';
 
 export interface CIStackProps extends StackProps {
   /** Should the Jenkins use https  */
@@ -200,6 +201,10 @@ export class CIStack extends Stack {
       listenerCertificate,
       useSsl,
       accessLogBucket: auditloggingS3Bucket.bucket,
+    });
+
+    const waf = new JenkinsWAF(this, {
+      loadBalancer: externalLoadBalancer.loadBalancer,
     });
 
     const artifactBucket = new Bucket(this, 'BuildBucket');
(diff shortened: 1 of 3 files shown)


prudhvigodithi (approved):

lib/ci-stack.ts
```diff
@@ -202,6 +203,10 @@ export class CIStack extends Stack {
       accessLogBucket: auditloggingS3Bucket.bucket,
     });
 
+    const waf = new JenkinsWAF(this, {
+      loadBalancer: externalLoadBalancer.loadBalancer,
```
peterzhuamazon, line 207: So this pointed to the Jenkins ALB? And will dynamically adjust if the ALB get replaced?

gaiksaya: Yes!

lib/security/waf.ts
```diff
@@ -0,0 +1,120 @@
...
+];
+
+export class WAF extends CfnWebACL {
+  constructor(scope: Construct, id: string) {
+    super(scope, id, {
+      defaultAction: { allow: {} },
+      visibilityConfig: {
+        cloudWatchMetricsEnabled: true,
+        metricName: 'jenkins-WAF',
+        sampledRequestsEnabled: false,
```
peterzhuamazon, line 89: Why is this one disabled here for sampled request?

peterzhuamazon (approved):

lib/security/waf.ts
gaiksaya: Good catch. Enabled it now. Please check

prudhvigodithi:
Also this `waf.node.addDependency(externalLoadBalancer)` should be added to avoid `waf` creating before the `externalLoadBalancer`
```
export class JenkinsWAF extends Stack {
  constructor(scope: Construct, id: string, props: WafProps) {
    const waf = new WAF(this, 'WAFv2');
    // Create an association with the alb
    new WebACLAssociation(this, 'wafALBassociation', {
      resourceArn: props.loadBalancer.loadBalancerArn,
      webAclArn: waf.attrArn,
    });
  }
}
```
Then 
```
    const waf = new JenkinsWAF(this, {
      loadBalancer: externalLoadBalancer.loadBalancer,
    });
    waf.node.addDependency(externalLoadBalancer)
```

gaiksaya:
CDK is smart enough to detect that @prudhvigodithi. It creates the dependency graph automatically for resources. Regarding stacks we need to specify.

PR #434, opened 2024-06-27
Add cloudwatch logs for gradle-check job

rishabh6788:
### Description
Add cloudwatch logs for gradle-check job. 
In the future similar configuration can be added if we want to onboard other job logs as well. 

### Issues Resolved

### DIFF
--- a/lib/compute/jenkins-main-node.ts
+++ b/lib/compute/jenkins-main-node.ts
@@ -393,6 +393,12 @@ export class JenkinsMainNode {
                   auto_removal: true,
                   log_stream_name: 'workflow-logs',
                 },
+                {
+                  file_path: '/var/lib/jenkins/jobs/gradle-check/builds/*/log',
+                  log_group_name: 'JenkinsMainNode/gradle-check.log',
+                  auto_removal: true,
+                  log_stream_name: 'gradle-check.log/{date}',
+                },
               ],
             },
           },


lib/compute/jenkins-main-node.ts
```diff
@@ -393,6 +393,12 @@ export class JenkinsMainNode {
                   auto_removal: true,
                   log_stream_name: 'workflow-logs',
                 },
+                {
+                  file_path: '/var/lib/jenkins/jobs/gradle-check/builds/*/log',
+                  log_group_name: 'JenkinsMainNode/gradle-check.log',
+                  auto_removal: true,
+                  log_stream_name: 'gradle-check.log/{date}',
```
gaiksaya, line 400: Would date create too many log-streams? Also forgot what auto-removal does. Any idea?

rishabh6788: There is no limit on how many log streams can be created. 
The choice is either push everything into one log stream or have one created daily so that it becomes easier to find the log one is looking for and faster, since it will contain only limited logs.

rishabh6788: Reg `auto_removal`, it is mainly helpful when you have some sort of log rotation happening, particularly on logs that are suffixed, e.g., sys.log is your main log file and your rotation logs creates sys.log.1 after each hour. In this case CW is supposed to clean up the rotated log. 

But in our case it is not the case and the flag doesn't do anything.

gaiksaya: Yeah we can think about adding log retention policy. I believe it will be just too much logs. Even jenkins has 30 day retention policy for gradle-check logs. Can come as an enhancement.

gaiksaya (approved):

Issue #509, opened 2024-11-19
Configure agent-node usage based on jenkins instance type


Divyaasm:
Current agent node setup and initialization is confined to use all the supported agent types while creating a jenkins instance as part of stack creation.

- Reorganize the agent node types into three different types by separating into agent types used by release/test/build workflows, benchmark test-workflows and gradle check workflows.
- We should be able bring up a requested jenkins instance type as mentioned above, only with the necessary agent nodes required by the workflows using it.

Divyaasm:
Closed via https://github.com/opensearch-project/opensearch-ci/pull/513

PR #513, opened 2024-12-05
Configure agent nodes based on jenkins instance type

Divyaasm:
### Description
Changes to include the agent nodes required for a jenkins instance type taking the values Bulid,Test,Release,  benchmark, gradle.
This implentation brings up all agent nodes for type BTR, future enhancement will only bring up B/T/R related agent nodes only 

### Issues Resolved
https://github.com/opensearch-project/opensearch-ci/issues/509


lib/ci-stack.ts
```diff
@@ -52,6 +52,8 @@ export interface CIStackProps extends StackProps {
   readonly enableViews?: boolean;
   /** Use Production Agents */
   readonly useProdAgents?: boolean;
+  /** Specify jenkins instance type */
+  readonly jenkinsType?: string;
```
rishabh6788, line 56: Nit: `jenkinsDeploymentType`?

lib/compute/agent-nodes.ts
```diff
@@ -281,4 +281,15 @@ export class AgentNodes {
       remoteFs: '/home/ec2-user',
     };
   }
+
+  public getRequiredAgentNodes(type: string): any[] {
+    return Object.keys(this)
+      .filter((key) => {
```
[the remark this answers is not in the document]
Divyaasm: We are doing similar thing with _jenkinsType_ variable.
If _jenkinsType_ value is not passed we assume it's default(this option is mostly used for testing), in all other cases we retrieve the agent nodes list with the value passed in _jenkinsType_(Ex: benchmark, gradle). But for BTR we need to exclude the agentnode names having the values ('default', 'benchmark', 'gradle').

The Drawback is agentnodes used by BTR instance cannot be differentiated from their names like we do with benchmark and gradle. We can think of updating agentnodes names used by BTR with a key_name(RELEASE) if it's straightforward.

lib/ci-stack.ts
Divyaasm: Or jenkinsInstanceType, Which name would you recommend?

rishabh6788: I'm good with both.

lib/compute/agent-nodes.ts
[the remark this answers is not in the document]
rishabh6788: I think since we have a defined number of instance type, better have a check, may be enums, to verify if a valid instance type name is passed. 
If invalid name is passed then throw an error.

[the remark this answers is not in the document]
gaiksaya: Agreed. Have enums for instance types. Also what is BTR? 😅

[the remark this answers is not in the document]
Divyaasm: Yes Sayali

rishabh6788 (approved):
LGTM!  @gaiksaya any other concerns?

lib/ci-stack.ts
```diff
@@ -28,6 +28,13 @@ import { JenkinsSecurityGroups } from './security/ci-security-groups';
 import { JenkinsWAF } from './security/waf';
 import { FineGrainedAccessSpecs } from './compute/auth-config';
 
+enum DeploymentType {
+  BTR='BTR',
```
gaiksaya, line 32: Just to clarify for external community.
```suggestion
  BTR='BTR', // Build Test Release
```

lib/ci-stack.ts
```diff
@@ -53,6 +60,8 @@ export interface CIStackProps extends StackProps {
   readonly enableViews?: boolean;
   /** Use Production Agents */
   readonly useProdAgents?: boolean;
+  /** Specify jenkins instance type */
+  readonly jenkinsInstanceType?: string;
```
gaiksaya, line 64: Let's keep it constant across the code. Either deploymentType or JenkinsInstanceType. 
Also maybe the type should be `string | DeploymentType`

lib/ci-stack.ts
```diff
@@ -149,34 +166,15 @@ export class CIStack extends Stack {
     const listenerCertificate = ListenerCertificate.fromArn(certificateArn.secretValue.toString());
     const agentNode = new AgentNodes(this);
 
-    if (useProdAgents.toString() === 'true') {
-      // eslint-disable-next-line no-console
-      console.warn('Please note that if you have decided to use the provided production jenkins agents then '
+    // eslint-disable-next-line no-console
+    console.warn('Please note that if you have decided to use the provided production jenkins agents then '
```
gaiksaya, line 170: Isn't the conditional `useProdAgents` required here?
The warning was only relevant if the ProdAgents parameter is enabled.

lib/compute/agent-node-config.ts
```diff
@@ -23,7 +23,7 @@ export interface AgentNodeProps {
   amiId: string;
   instanceType: string;
   customDeviceMapping: string;
-  workerLabelString: string;
+  workerLabelString: string[];
```
gaiksaya, line 26: How about make it `[string, DeploymentType]`. This way it will by default only accept the specified enum values.

Divyaasm: We already have a check for the enum values
> 
> else if (!Object.values(DeploymentType).includes(jenkinsInstanceType as DeploymentType)) {
>       throw new Error(`Invalid jenkinsInstanceType value: ${jenkinsInstanceType}`);

and we have hardcoded the DeploymentType for each agent property anyway

> workerLabelString: ['Jenkins-Agent-AL2023-X64-C54xlarge-Single-Host', 'BTR'],  

Do we still require to change the type @gaiksaya?

Divyaasm: Addressed all other comments in the latest commit @gaiksaya

gaiksaya: The condition will only be checked if the given value is undefined as per this code: https://github.com/opensearch-project/opensearch-ci/pull/513/files#diff-4b6fe4122d26e45b42b65d74533746571fed0807fe76bb2df2ea33b75227e004R130-R134
We want to check despite of that

gaiksaya: Synced up offline. The if else block does the needful. We can iterate on this logic later.

gaiksaya (approved):

PR #514, opened 2024-12-06
Add support for fine-grained access control

### LINKED ISSUE #507: Enable fine grain access per user per workflow or set of workflows

Coming from https://github.com/opensearch-project/opensearch-build/issues/5171
The current jenkins infrastructure has permissions set up which allows only admin or readonly access. See https://github.com/opensearch-project/opensearch-ci/blob/main/lib/compute/oidc-config.ts#L82-L85

Have more fine grain access such that:
- User A can access:
    - Workflow A
    - Workflow B
    - Workflow C
 - User B can access:
     - Workflow A
     - Everything under folder called `foo`
 
- Default read-only and admin access should be retained as is. 
- Everything including user handle, workflow names or pattern or folder and exact permission should be infrastructure as code.

### Describe alternatives you've considered

_No response_

### Additional context

[Role-based Authorization Strategy](https://plugins.jenkins.io/role-strategy/) plugin and [config](https://github.com/opensearch-project/opensearch-ci/blob/main/lib/compute/oidc-config.ts#L79) is already being used. Enhance on top of it to give fine grain access.

gaiksaya:
### Description
This change adds support for fine grained access control. Access can now be controlled at workflow levels per user.
Also fixes the test cases where earlier an already committed Yaml was being parsed. However, we need to actually parse the YAML object. 
Deletes the unused test data.

### Issues Resolved
#507 


lib/compute/auth-config.ts
```diff
@@ -103,14 +125,20 @@ export class AuthConfig {
...
           },
-
           ],
         },
       },
     };
 
     jenkinsYaml.jenkins.authorizationStrategy = rolesAndPermissions;
-
+    if (typeof fineGrainedAccessItems !== 'undefined') {
```
peterzhuamazon, line 134: Does `!== undefined` better than `=== true` or similar?

lib/compute/jenkins-main-node.ts
```diff
@@ -452,7 +453,8 @@ export class JenkinsMainNode {
     agentNodeObject: AgentNodeConfig, props: AgentNodeNetworkProps, agentNode: AgentNodeProps[], macAgent: string): string {
     let updatedConfig = agentNodeObject.addAgentConfigToJenkinsYaml(stack, agentNode, props, macAgent);
     if (loginAuthProps.authType !== 'default') {
-      updatedConfig = AuthConfig.addOidcConfigToJenkinsYaml(updatedConfig, loginAuthProps.authType, loginAuthProps.adminUsers);
+      updatedConfig = AuthConfig.addOidcConfigToJenkinsYaml(updatedConfig, loginAuthProps.authType,
+        loginAuthProps.adminUsers, loginAuthProps.fineGrainedAccessSpecs);
```
peterzhuamazon, line 457: Will admins have all the permissions defined in above?
And what about read only users?

lib/compute/auth-config.ts
gaiksaya: Both have different purpose. One checks if value is undefined and other one checks if value is truly assigned to the variable.

lib/compute/jenkins-main-node.ts
gaiksaya: Admin users will have admin permissions. Same applies for readonly users. The roles are divided into 2 types. global and items at jenkins level. Admin and readonly falls under `global` and fineGrainedAccess falls under `items` category.

rishabh6788:
Can you please explain if I want to add a new user and give them build acess to a particular job, how would I achieve that?

gaiksaya:
> Can you please explain if I want to add a new user and give them build acess to a particular job, how would I achieve that?

Like this https://github.com/opensearch-project/opensearch-ci/pull/514/files#diff-26e449f9955e460d5aabdf24a751a58d9aaace1a73226f143618601a18d0af32R52-R57
The regex pattern defines the access to the workflows. For example:
`((distribution|integ).*)` will give access to all jobs that start with `distrbution-*` and `integ-*`

peterzhuamazon (approved):

peterzhuamazon:
Asking @rishabh6788 to do another pass before merging.

Thanks!

rishabh6788 (approved):

Issue #515, opened 2024-12-19
[Bug]: OIDC config is invalid after OIDC plugin breaking change


lewijacn:
### Describe the bug

The current non-github OIDC auth config is constructed similar to: 
```
{
     "clientId": "example_id",
     "clientSecret": "example_password",
     "wellKnownOpenIDConfigurationUrl": "https://www.example.com",
     "tokenServerUrl": "https://example.com/token",
     "authorizationServerUrl": "https://example.com/authorize",
     "userInfoServerUrl": "https://example.com/userinfo"
 }
```

However after breaking changes introduced in `4.350` of the OIDC plugin, this format is no longer valid for CASC usage and will produce errors similar to below:
```
io.jenkins.plugins.casc.UnknownAttributesException: oic: Invalid configuration elements for type: class org.jenkinsci.plugins.oic.OicSecurityRealm : authorizationServerUrl,userInfoServerUrl,tokenServerUrl,wellKnownOpenIDConfigurationUrl,scopes.
Available attributes : allowTokenAccessWithoutOicSession, allowedTokenExpirationClockSkewSeconds, captchaSupport, clientId, clientSecret, disableSslVerification, disableTokenVerification, emailFieldName, escapeHatchEnabled, escapeHatchGroup, escapeHatchSecret, escapeHatchUsername, fullNameFieldName, groupsFieldName, logoutFromOpenidProvider, nonceDisabled, pkceEnabled, postLogoutRedirectUrl, rootURLFromRequest, sendScopesInTokenRequest, serverConfiguration, tokenExpirationCheckDisabled, tokenFieldToCheckKey, tokenFieldToCheckValue, userNameField
```

Reference: https://github.com/jenkinsci/oic-auth-plugin/releases

### To reproduce

To reproduce a clean deployment with `authType=oidc` can be used and logs will be produced indicating this invalid format

### Expected behavior

Expected format should be similar to:
```
{
     "clientId": "example_id",
     "clientSecret": "example_password",
     "serverConfiguration": {
        "wellKnown": {
          "wellKnownOpenIDConfigurationUrl": "https://example.com/.well-known/openid-configuration"
     }
 }
```
Or below depending on use case:
```
{
     "clientId": "example_id",
     "clientSecret": "example_password",
     "serverConfiguration": {
         "manual": {
             "issuer": "https://example.com/api/oauth2/v2"
             "tokenServerUrl": "https://example.com/token",
[...]

PR #516, opened 2025-01-08
Add config helper utility to improve modifying Jenkins CASC yaml

lewijacn:
### Description
This feature adds a `configHelper` python utility script to make modifying the Jenkins CASC yaml easier to understand and test locally. This script replaces previous shell functionality to add authConfig into the `jenkins.yaml` and extends support to allow nested secret structures to be placed in the authConfig secret which unblocks the current OIC auth pattern which has had a breaking change that requires this nested structure.

### Issues Resolved
https://github.com/opensearch-project/opensearch-ci/issues/515

### DIFF
--- a/configHelper/configHelper/config_helper.py
+++ b/configHelper/configHelper/config_helper.py
@@ -46,7 +46,7 @@ def main():
     parser = argparse.ArgumentParser(description="CI config setup helper.")
     parser.add_argument("--initial-jenkins-config-file-path", type=str, required=True, help="The file path for the initial jenkins config yaml file to load from and be modified in place, e.g. /initial_jenkins.yaml")
     parser.add_argument("--auth-aws-secret-arn", type=str, required=True, help="The AWS Secrets Manager Secret ARN to pull auth config from and populate into jenkins config")
-    parser.add_argument("--aws-region", type=str, required=True, help="The AWS region for the boto3 client to use")
+    parser.add_argument("--aws-region", type=str, required=True, help="The AWS region for the boto3 client to use, e.g. us-east-1")
     parser.add_argument("--auth-type", type=str, required=True, help="The Jenkins auth type to use", choices=['oidc', 'github'])
     args = parser.parse_args()
     file_path = args.initial_jenkins_config_file_path
--- a/configHelper/configHelper/config_helper.py
+++ b/configHelper/configHelper/config_helper.py
@@ -35,9 +35,13 @@ def get_secret_as_dict(secret_arn, aws_region):
 def modify_auth_config(yaml_dict, auth_aws_secret_arn, aws_region, auth_type):
     security_realm_id = "oic" if auth_type.lower() == "oidc" else auth_type.lower()
     auth_secret_dict = get_secret_as_dict(auth_aws_secret_arn, aws_region)
+    if "securityRealm" not in yaml_dict.get("jenkins", {}):
+        logging.warning('Unable to find jenkins.securityRealm path in initial Jenkins config yaml, creating securityRealm object')
+        yaml_dict.get("jenkins", {})["securityRealm"] = {}
+    if security_realm_id not in yaml_dict.get("jenkins", {}).get("securityRealm", {}):
+        logging.warning(f'Unable to find jenkins.securityRealm.{security_realm_id} path in initial Jenkins config yaml, creating {security_realm_id} object')
+        yaml_dict.get("jenkins", {}).get("securityRealm", {})[security_realm_id] = {}
     yaml_auth_dict = yaml_dict.get("jenkins", {}).get("securityRealm", {}).get(security_realm_id, {})
-    if not yaml_auth_dict:
-        raise RuntimeError(f'Unable to find jenkins.securityRealm.{security_realm_id} path in initial Jenkins config yaml')
     merge_dicts(yaml_auth_dict, auth_secret_dict)
 
 
--- a/configHelper/configHelper/config_helper.py
+++ b/configHelper/configHelper/config_helper.py
@@ -1,0 +1,76 @@
+import argparse
+import boto3
+import logging
+import json
+import yaml
+
+logging.basicConfig(format='%(asctime)s [%(levelname)s] %(message)s', level=logging.INFO)
+logger = logging.getLogger(__name__)
+
+
+def merge_dicts(dict1, dict2):
+    """Recursively merges two dictionaries."""
+    for key, value in dict2.items():
+        if key in dict1 and isinstance(dict1[key], dict) and isinstance(value, dict):
+            merge_dicts(dict1[key], value)
+        else:
+            dict1[key] = value
+
+def get_secret_as_dict(secret_arn, aws_region):
+    client = boto3.client('secretsmanager', region_name=aws_region)
+    response = client.get_secret_value(SecretId=secret_arn)
+
+    if 'SecretString' in response:
+        secret_string = response['SecretString']
+    else:
+        raise RuntimeError(f'SecretString not found in response when retrieving secret value: {secret_arn}')
+
+    try:
+        secret_dict = json.loads(secret_string)
+    except json.JSONDecodeError as json_error:
+        raise RuntimeError(f"Failed to parse secret string as JSON: {json_error}")
+    return secret_dict
+
+
+def modify_auth_config(yaml_dict, auth_secret_arn, aws_region, security_realm_id):
+    auth_secret_dict = get_secret_as_dict(auth_secret_arn, aws_region)
+    yaml_auth_dict = yaml_dict.get("jenkins", {}).get("securityRealm", {}).get(security_realm_id, {})
+    if not yaml_auth_dict:
+        raise RuntimeError(f'Unable to find jenkins.securityRealm.{security_realm_id} path in initial Jenkins config yaml')
+    merge_dicts(yaml_auth_dict, auth_secret_dict)
+
+
+def main():
+    logging.info("Starting process to load auth config into Jenkins config yaml")
+    parser = argparse.ArgumentParser(description="CI config setup helper.")
+    parser.add_argument("--jenkins-config-file-path", type=str, required=True, help="The file path for the jenkins config yaml file, e.g. /initial_jenkins.yaml")
+    parser.add_argument("--auth-secret-arn", type=str, required=True, help="The AWS Secrets Manager Secret ARN to pull auth config from and populate into jenkins config")
+    parser.add_argument("--aws-region", type=str, required=True, help="The AWS region for the boto3 client to use")
+    parser.add_argument("--security-realm-id", type=str, required=True, help="The Jenkins security realm id to use, e.g. oic,github")
+    args = parser.parse_args()
+    file_path = args.jenkins_config_file_path
+    logging.info(f"The following args were provided: {args}")
+
+    # Read input file data
+    try:
+        with open(file_path, 'r') as file:
+            yaml_data = yaml.safe_load(file)
+    except FileNotFoundError:
+        raise FileNotFoundError(f"Error: The file '{file_path}' does not exist.")
+    except yaml.YAMLError as e:
+        raise ValueError(f"Error parsing YAML file '{file_path}': {e}")
+
+    # Modify input data as needed
+    modify_auth_config(yaml_data, args.auth_secret_arn, args.aws_region, args.security_realm_id)
+
+    # Write file with updated data
+    try:
+        with open(file_path, 'w') as file:
+            yaml.dump(yaml_data, file, default_flow_style=False)
+    except Exception as e:
+        raise IOError(f"Error writing to file '{file_path}': {e}")
+    logging.info("Process completed to load auth config into Jenkins config yaml")
+
+
+if __name__ == "__main__":
+    main()
--- a/configHelper/configHelper/config_helper.py
+++ b/configHelper/configHelper/config_helper.py
@@ -32,8 +32,9 @@ def get_secret_as_dict(secret_arn, aws_region):
     return secret_dict
 
 
-def modify_auth_config(yaml_dict, auth_secret_arn, aws_region, security_realm_id):
-    auth_secret_dict = get_secret_as_dict(auth_secret_arn, aws_region)
+def modify_auth_config(yaml_dict, auth_aws_secret_arn, aws_region, auth_type):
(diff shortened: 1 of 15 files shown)


configHelper/configHelper/config_helper.py
```diff
@@ -0,0 +1,76 @@
...
+    except json.JSONDecodeError as json_error:
+        raise RuntimeError(f"Failed to parse secret string as JSON: {json_error}")
+    return secret_dict
+
+
+def modify_auth_config(yaml_dict, auth_secret_arn, aws_region, security_realm_id):
+    auth_secret_dict = get_secret_as_dict(auth_secret_arn, aws_region)
+    yaml_auth_dict = yaml_dict.get("jenkins", {}).get("securityRealm", {}).get(security_realm_id, {})
+    if not yaml_auth_dict:
+        raise RuntimeError(f'Unable to find jenkins.securityRealm.{security_realm_id} path in initial Jenkins config yaml')
```
gaiksaya, line 39: Sometimes if security is disabled, I believe securityRealm can be absent (not sure about this but I remember deleting everything security to disable security in my dev set up), in that case do we want to add the securityRealm instead of throwing an error?
WDYT?

configHelper/tests/resources/sampleJenkinsOICAuth.yaml
```diff
@@ -0,0 +1,256 @@
...
+  myViewsTabBar: "standard"
+  numExecutors: 4
+  primaryView:
+    all:
+      name: "all"
+  projectNamingStrategy: "standard"
+  quietPeriod: 5
+  scmCheckoutRetryCount: 0
+  securityRealm:
+    oic:
```
gaiksaya, line 33: Not sure why they wrote `oic` instead of `oidc` in the config, how do you think this should be handled while adding the config. Let user add `oic` instead of `oidc` or keep the `oidc` norm and let the code handle the conversion?

gaiksaya (commented):
Thanks for the contribution @lewijacn. This looks promising and relatively safe compared to previous implementation 🙌

configHelper/configHelper/config_helper.py
```diff
@@ -0,0 +1,76 @@
...
+    merge_dicts(yaml_auth_dict, auth_secret_dict)
+
+
+def main():
+    logging.info("Starting process to load auth config into Jenkins config yaml")
+    parser = argparse.ArgumentParser(description="CI config setup helper.")
+    parser.add_argument("--jenkins-config-file-path", type=str, required=True, help="The file path for the jenkins config yaml file, e.g. /initial_jenkins.yaml")
+    parser.add_argument("--auth-secret-arn", type=str, required=True, help="The AWS Secrets Manager Secret ARN to pull auth config from and populate into jenkins config")
```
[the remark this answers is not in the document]
lewijacn: My intention was for this python script to be extended to potentially cover other things we do with `InitCommand` and thus possibly pull in other secrets so I'd like to keep it somewhat defined, I'll be a bit more descriptive with `auth-aws-secret-arn` lmk if you have any concern

lib/compute/jenkins-main-node.ts
```diff
@@ -430,16 +431,14 @@ export class JenkinsMainNode {
...
       InitCommand.shellCommand(loginAuthProps.authType !== 'default'
         // eslint-disable-next-line max-len
-        ? `var=\`aws --region ${stackRegion} secretsmanager get-secret-value --secret-id ${loginAuthProps.authCredsSecretsArn} --query SecretString --output text\` && `
-        + ' varkeys=`echo $var | yq \'keys\' | cut -d "-" -f2 | cut -d " " -f2` &&'
-        // eslint-disable-next-line max-len
-        + ` for i in $varkeys; do newvalue=\`echo $var | yq .$i\` && myenv=$newvalue i=$i yq -i '.jenkins.securityRealm.${realm}.[env(i)]=env(myenv)' /initial_jenkins.yaml ; done`
-        : 'echo No changes made to initial_jenkins.yaml with respect to OIDC'),
-
+        ? `cd /configHelper && sudo pipenv run python3 config_helper.py --jenkins-config-file-path=/initial_jenkins.yaml --auth-secret-arn=${loginAuthProps.authCredsSecretsArn} --security-realm-id=${realm} --aws-region=${stackRegion} > configHelper.log 2>&1`
+        : 'No changes made to initial_jenkins.yaml with respect to OIDC'),
```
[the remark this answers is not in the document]
lewijacn: Committed

configHelper/configHelper/config_helper.py
lewijacn: Currently we do not call this python script unless an auth type is given. We definitely could alter the python script to be able to handle say a default auth type and essentially just print a log line. I'm fine with either approach, do you have any preference? 
```
      // Make any changes to initial jenkins.yaml
      InitCommand.shellCommand(loginAuthProps.authType !== 'default'
        ? `cd /configHelper && sudo pipenv run python3 config_helper.py --jenkins-config-file-path=/initial_jenkins.yaml --auth-secret-arn=${loginAuthProps.authCredsSecretsArn} --security-realm-id=${realm} --aws-region=${stackRegion} > configHelper.log 2>&1`
        : 'No changes made to initial_jenkins.yaml with respect to AuthType'),
```

configHelper/tests/resources/sampleJenkinsOICAuth.yaml
lewijacn: I think from my perspective this is pretty abstracted away from users and what we have should be fine. The user would only be providing values within this `oic` block and unless a user is digging around their Jenkins yaml they they shouldn't be aware, and even then it should be a small dot to connect

configHelper/configHelper/config_helper.py
```diff
@@ -0,0 +1,76 @@
...
+    parser.add_argument("--aws-region", type=str, required=True, help="The AWS region for the boto3 client to use")
+    parser.add_argument("--security-realm-id", type=str, required=True, help="The Jenkins security realm id to use, e.g. oic,github")
```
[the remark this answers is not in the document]
lewijacn: Yeah that's a good call-out. Let me make this `auth-type` and have the options `oidc, github`. See my other comment if you want to also start calling this script with `default` in which case we can add that as a choice as well

configHelper/configHelper/config_helper.py
```diff
@@ -0,0 +1,76 @@
...
+    yaml_auth_dict = yaml_dict.get("jenkins", {}).get("securityRealm", {}).get(security_realm_id, {})
+    if not yaml_auth_dict:
+        raise RuntimeError(f'Unable to find jenkins.securityRealm.{security_realm_id} path in initial Jenkins config yaml')
+    merge_dicts(yaml_auth_dict, auth_secret_dict)
+
+
+def main():
+    logging.info("Starting process to load auth config into Jenkins config yaml")
+    parser = argparse.ArgumentParser(description="CI config setup helper.")
+    parser.add_argument("--jenkins-config-file-path", type=str, required=True, help="The file path for the jenkins config yaml file, e.g. /initial_jenkins.yaml")
```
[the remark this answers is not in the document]
lewijacn: I don't have much preference here and can change to this. We do change this file in place so the name I thought could potentially be confusing. I can also add to the description here to make that clearer, which I will do

lewijacn:
> Thanks for the contribution @lewijacn. This looks promising and relatively safe compared to previous implementation 🙌

Thanks for the review, hopefully this will help make an easier and more testable path toward some Jenkins setup stuff that we can leverage in the future.

As an aside, I think we can run the python tests pretty easily in a github action as a small follow-up

configHelper/tests/resources/sampleJenkinsOICAuth.yaml
lewijacn: I think I see what you meant now, my preference is keep the oidc norm and let code handle the conversion

configHelper/configHelper/config_helper.py
```diff
@@ -0,0 +1,77 @@
...
+        raise RuntimeError(f'Unable to find jenkins.securityRealm.{security_realm_id} path in initial Jenkins config yaml')
+    merge_dicts(yaml_auth_dict, auth_secret_dict)
+
+
+def main():
+    logging.info("Starting process to load auth config into Jenkins config yaml")
+    parser = argparse.ArgumentParser(description="CI config setup helper.")
+    parser.add_argument("--initial-jenkins-config-file-path", type=str, required=True, help="The file path for the initial jenkins config yaml file to load from and be modified in place, e.g. /initial_jenkins.yaml")
+    parser.add_argument("--auth-aws-secret-arn", type=str, required=True, help="The AWS Secrets Manager Secret ARN to pull auth config from and populate into jenkins config")
+    parser.add_argument("--aws-region", type=str, required=True, help="The AWS region for the boto3 client to use")
```
rishabh6788, line 49: Nit: type would be helpful, `IAD` or `us-east-1`?

configHelper/configHelper/config_helper.py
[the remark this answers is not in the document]
gaiksaya: Makes sense to only use oidc and github if the script is only being called for non-default values. We can skip default.

configHelper/configHelper/config_helper.py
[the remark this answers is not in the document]
gaiksaya: Just for you reference, we are loading an initial jenkins.yaml file (from resources) and editing in place by parsing the yaml.

configHelper/configHelper/config_helper.py
gaiksaya: I agree that the script is only called when auth-type is specified. Was just concerned about the edge case where the initial jenkins.yaml might be missing `securityRealm` key altogether. We can get this in and monitor how it goes.

lewijacn: My expectation is that the python script would throw error and result in the `initial_jenkins.yaml` not getting updated. Curious if there is any particular way you want to handle this?

configHelper/configHelper/config_helper.py
lewijacn: Thanks, updated now

configHelper/configHelper/config_helper.py
gaiksaya: How about add the missing key `jenkins.securityRealm` if not found?

lewijacn: Pushed an update so that if we can't find the `jenkins.securityRealm` or auth object after it, we will create it

lib/compute/auth-config.ts
```diff
@@ -81,16 +81,11 @@ export class AuthConfig {
       oic: {
         clientId: 'clientId',
         clientSecret: 'clientSecret',
-        authorizationServerUrl: 'http://localhost',
-        wellKnownOpenIDConfigurationUrl: 'wellKnownOpenIDConfigurationUrl',
-        tokenServerUrl: 'tokenServerUrl',
-        userInfoServerUrl: 'userInfoServerUrl',
```
gaiksaya, line 87: Are these values not needed anymore?

lewijacn: They are no longer available with the updated plugin, error from the github issue: 
```
io.jenkins.plugins.casc.UnknownAttributesException: oic: Invalid configuration elements for type: class org.jenkinsci.plugins.oic.OicSecurityRealm : authorizationServerUrl,userInfoServerUrl,tokenServerUrl,wellKnownOpenIDConfigurationUrl,scopes.
Available attributes : allowTokenAccessWithoutOicSession, allowedTokenExpirationClockSkewSeconds, captchaSupport, clientId, clientSecret, disableSslVerification, disableTokenVerification, emailFieldName, escapeHatchEnabled, escapeHatchGroup, escapeHatchSecret, escapeHatchUsername, fullNameFieldName, groupsFieldName, logoutFromOpenidProvider, nonceDisabled, pkceEnabled, postLogoutRedirectUrl, rootURLFromRequest, sendScopesInTokenRequest, serverConfiguration, tokenExpirationCheckDisabled, tokenFieldToCheckKey, tokenFieldToCheckValue, userNameField
```

gaiksaya: I see. Thanks! In that case, I believe we need to add `serverConfiguration` as mentioned in the issue #515 right? 
Please correct me if I am wrong.

lewijacn: I didn't create this field as it would be just an empty object, and thought would be confusing. As long as the secret has the serverConfiguration field it will get added in:
   ```
    {
        "clientId": "example_id",
        "clientSecret": "example_password",
        "serverConfiguration": {
            "wellKnown": {
                "wellKnownOpenIDConfigurationUrl": "https://example.openid.com/.well-known/openid-configuration"
            }
        }
    }
    ```

gaiksaya: Ah okay! Since we are blinding merging the dictionaries.  However, if this is a required config for `oic`, do you think adding a sample values just like above would be helpful?

lewijacn: I considered this at one point, but I wanted to be able to support both configurations below, and with the current design if we use the example above and someone specifies a manual server configuration the objects would get joined together and likely cause issue
```
{
     "clientId": "example_id",
     "clientSecret": "example_password",
     "serverConfiguration": {
        "wellKnown": {
          "wellKnownOpenIDConfigurationUrl": "https://example.com/.well-known/openid-configuration"
     }
 }
OR
{
     "clientId": "example_id",
     "clientSecret": "example_password",
     "serverConfiguration": {
         "manual": {
             "issuer": "https://example.com/api/oauth2/v2"
             "tokenServerUrl": "https://example.com/token",
             "authorizationServerUrl": "https://example.com/authorize",
             "scopes": "open-id"
         }
     }
 }
 ```

Comment (gaiksaya) [reply lib/compute/auth-config.ts:87]: Oh I just meant `serverConfiguration` the nested values (manual or wellKnown) can be ignored.

Comment (lewijacn) [reply lib/compute/auth-config.ts:87]: Sure I can this as an empty field in the auth-config.ts file like so:
```
serverConfiguration: {},
```

lewijacn: Just pushed an update for this and verified with test

gaiksaya (approved):
LGTM! Thanks @lewijacn

gaiksaya:
@rishabh6788 @Divyaasm Any concerns before I merge in an hour?

PR #520, opened 2025-01-28
Add few fixes to minimize stack resource changes

Divyaasm:
### Description
Specifying availiability zones to avoid vpc re-creation
using -p to avoid folder already /var/lib/jenkins exists error! when dataRetention parameter is set to true

### Issues Resolved
https://github.com/opensearch-project/opensearch-ci/issues/517

### DIFF
--- a/lib/ci-stack.ts
+++ b/lib/ci-stack.ts
@@ -96,6 +96,7 @@ export class CIStack extends Stack {
 
     const auditloggingS3Bucket = new CiAuditLogging(this);
     const vpc = new Vpc(this, 'JenkinsVPC', {
+      availabilityZones: ['us-east-1a', 'us-east-1b', 'us-east-1c'],
       flowLogs: {
         s3: {
           destination: FlowLogDestination.toS3(auditloggingS3Bucket.bucket, 'vpcFlowLogs'),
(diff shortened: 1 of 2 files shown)


peterzhuamazon (approved):

lib/ci-stack.ts
```diff
@@ -96,6 +96,7 @@ export class CIStack extends Stack {
 
     const auditloggingS3Bucket = new CiAuditLogging(this);
     const vpc = new Vpc(this, 'JenkinsVPC', {
+      availabilityZones: ['us-east-1a', 'us-east-1b', 'us-east-1c'],
```
gaiksaya, line 99: This should be dynamic. I always use this repo to deploy in us-west-2. In that case, the deployment would fail. Either take this as context parameters or keep whatever is default.

gaiksaya (changes requested):

lib/ci-stack.ts
Divyaasm: Without this change VPC is getting recreated. Hardcodes as both prod and beta deploy in us-east-1. Passing the values thorough context parameter can be implemented.

gaiksaya: Yeah that is just for https://build.ci.opensearch.org/. Take into consideration for users deploying in their own dev environment.

Issue #527, opened 2025-02-10
Enhance the mechanism to overcome the usage of stack context parameters


Divyaasm:
### Is your feature request related to a problem? Please describe

We are currently passing separately context parameters while deploying stacks. With time we will be having more values which might increase maintenance. Have a better approach to pass the required stack parameters especially to the deployment yaml Workflows. 

### Describe the solution you'd like

Use a json file to store the stack parameter values for beta and prod environments, retrieve the parameter values from the json file while deploying stacks through yaml workflow.

PR #531, opened 2025-02-24
Create a deployment ci-stack for prod and beta deployment.

Divyaasm:
### Description
Add specific deployment way for beta and prod deployments. Will be referenced through github yaml deployment workflows

### Issues Resolved
https://github.com/opensearch-project/opensearch-ci/issues/527


bin/deployments/ci-stack.ts
```diff
@@ -0,0 +1,60 @@
...
+  ignoreResourcesFailures: fileConfig.ignoreResourcesFailures,
+  adminUsers: fileConfig.adminUsers,
+  dataRetention: fileConfig.dataRetention,
+  agentAssumeRole: StageDef.agentAssumeRole,
+  macAgent: fileConfig.macAgent,
+  restrictServerAccessTo: isProd ? Peer.anyIpv4() : Peer.prefixList('pl-60b85b09'),
+  useProdAgents: fileConfig.useProdAgents,
+  enableViews: fileConfig.enableViews,
+  fineGrainedAccessSpecs: [benchmarkFineGrainAccess],
+  envVarsFilePath: './resources/envVars.yaml',
```
gaiksaya, line 54: Why parse another file for these values. We can hard code these values as per our use case.

gaiksaya (changes requested):
Thanks for the change @Divyaasm 

Couple of questions:
I dont understand the flow. What is the importance of config.json here?
Why not add 2 stages om stage-definitions and iterate both of them in deployments/ci-stack.ts

Also for all the values we can hard code the properties. See the comment in ci-stack.ts. We have a generic version of stacks in bin/ci-stack.ts. No need to maintain another generic version. Also please consider moving the deployments folder to the repo root instead of bin. 

For external user who do not care about deployments need not understand or know about it.

Divyaasm:
> Thanks for the change @Divyaasm
> 
> Couple of questions: I dont understand the flow. What is the importance of config.json here? Why not add 2 stages om stage-definitions and iterate both of them in deployments/ci-stack.ts
> 
> Also for all the values we can hard code the properties. See the comment in ci-stack.ts. We have a generic version of stacks in bin/ci-stack.ts. No need to maintain another generic version. Also please consider moving the deployments folder to the repo root instead of bin.
> 
> For external user who do not care about deployments need not understand or know about it.

Hey @gaiksaya ,we have all the context parameters that can be hardcoded in the config.json file. Kept it separate for easy understandability. 
And we however deploy the beta and prod stacks seperately, which will need dynamic values for beta and prod and has to be stored as a secret. So the stacks can be generated based on the environment type Beta or Prod only when require throgh gh actions workflow.

To have the stacks on root folder. This will need additional files like package.json etc. constrained to stacks used in deployment folder as the target stack file path changes. Having it in bin folder, there won't any path clashes.

PR #535, opened 2025-03-13
Add additonal WAF rules to load balancer

zelinh:
### Description
Add additonal WAF rules to load balancer

### Issues Resolved


lib/security/waf.ts
```diff
@@ -77,6 +107,52 @@ const awsManagedRules: WafRule[] = [
       },
     },
   },
+  // AWS Managed Rules Known Bad Inputs Rule Set
+  {
+    name: 'AWS-AWSManagedRulesKnownBadInputsRuleSet',
+    rule: {
+      name: 'AWS-AWSManagedRulesKnownBadInputsRuleSet',
+      priority: 4,
```
gaiksaya, line 115: This needs to be priority 2

gaiksaya (commented):
Hi @zelinh 
Thanks for creating the PR. 
We need to allow list the GitHub IPs in the same PR else those will be blocked and gradle-checks as well as releases will start failing. Can you please add those as well?

lib/security/waf.ts
zelinh: Updated the PR with new rules. I put allowed GitHub Ip the highest priority to avoid block by other AWS managed rules in case.

lib/security/allowed_github_ips.json
```diff
@@ -0,0 +1,4999 @@
+{
+  "GitHubAllowIpIpv4": [
```
gaiksaya, line 2: We already have file named with allowed_github_ips.
```suggestion
  "ipv4": [
```

lib/security/allowed_github_ips.json
```diff
@@ -0,0 +1,4999 @@
...
+    "213.199.180.32/28",
+    "213.199.180.96/27",
+    "213.199.180.192/27",
+    "213.199.183.0/24",
+    "192.30.252.0/22",
+    "185.199.108.0/22",
+    "140.82.112.0/20",
+    "143.55.64.0/20"
+  ],
+  "GitHubAllowIpIpv6": [
```
gaiksaya, line 4090:
```suggestion
  "ipv6": [
```

lib/security/waf.ts
```diff
@@ -110,7 +199,68 @@ export interface WafProps extends StackProps{
 
 export class JenkinsWAF {
   constructor(stack: Stack, props: WafProps) {
-    const waf = new WAF(stack, 'WAFv2');
+    // Create IP Sets for IPv4 and IPv6
+    const ipSetIpv4 = new CfnIPSet(stack, 'GitHubIpv4Set', {
+      name: 'GitHubIpv4Set',
```
gaiksaya, line 204:
```suggestion
      name: 'AllowGitHubIpv4Set',
```

lib/security/waf.ts
```diff
@@ -110,7 +199,68 @@ export interface WafProps extends StackProps{
...
+      ipAddressVersion: 'IPV4',
+      scope: 'REGIONAL',
+      addresses: allowedIps.GitHubAllowIpIpv4,
+    });
+
+    const ipSetIpv6 = new CfnIPSet(stack, 'GitHubIpv6Set', {
+      name: 'GitHubIpv6Set',
```
gaiksaya, line 211:
```suggestion
      name: 'AllowGitHubIpv6Set',
```

gaiksaya (commented):
Another suggestion is using regex for ipv4 and ipv6 to parse the ips on the go if we do not want to hard code it. 
Adding @rishabh6788 to check the priority list.

lib/security/allowed_github_ips.json
zelinh: Updated.

lib/security/allowed_github_ips.json
zelinh: Updated.

lib/security/waf.ts
zelinh: This is just the IP set name and I don't think we should imply the action on these sets. We create the rule named `AllowGitHubIPv4`.

gaiksaya (approved):
@rishabh6788 Can you please check priorities on this one since you have more knowledge about it? Just making sure we are not deprioritizing important ones.

rishabh6788 (approved):
LGTM

PR #551, opened 2025-04-14
Add ci stack deployment workflows

rishabh6788:
### Description
This PR adds a deployment workflow to deploy ci stacks in beta and production environment. 

### Issues Resolved


.github/workflows/deploy-stacks.yml
```diff
@@ -0,0 +1,117 @@
...
+
+      - name: Setup Node and install cdk cli
+        uses: actions/setup-node@v4
+        with:
+          node-version: 20.x
+
+      - name: Install dependencies and build project
+        run: |
+          npm install
+          npm install -g aws-cdk
```
gaiksaya, line 26: Do we need to specify the version just to be sure not to break the compatibility with the current code-base?

.github/workflows/deploy-stacks.yml
```diff
@@ -0,0 +1,117 @@
...
+          npm run build
+
+      - name: Set up Env variables
+        run: |
+          echo "BETA_ACCOUNT_ID=${{ secrets.BETA_ACCOUNT_ID }}" >> $GITHUB_ENV
+          echo "PROD_ACCOUNT_ID=${{ secrets.PROD_ACCOUNT_ID }}" >> $GITHUB_ENV
+          echo "ASSUMED_ROLES=${{ secrets.ASSUMED_ROLES }}" >> $GITHUB_ENV
+          echo "REGION=${{ secrets.REGION }}" >> $GITHUB_ENV
```
gaiksaya, line 34: REGION does not need to be a secret right? Nothing to hide in there.

.github/workflows/deploy-stacks.yml
```diff
@@ -0,0 +1,117 @@
...
+              github.rest.repos.listCollaborators,
+              {
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                affiliation: 'all',
+                per_page: 100
+              }
+            );
+            const maintainers = collaborators.filter(c => c.permissions.maintain).map(c => c.login);
+            return maintainers;
```
gaiksaya, line 52: This would include everyone that has write access to the repository right? Or is is just maintain? 
Would it make more sense to parse the codeowners or maintainer file instead?

.github/workflows/deploy-stacks.yml
```diff
@@ -0,0 +1,117 @@
...
+
+            ```
+            ${{ steps.beta_stack_diff.outputs.diff_output }}
+            ```
+            </details>
+      - name: Deploy beta stacks
+        run: |
+          cdk deploy "*" -c stage=Beta --require-approval=never
+
+      - name: Configure Production deployment credentials
```
gaiksaya, line 87: Do we need bake time? Maybe just 5-10 min just to make sure maintainers/admins of jenkins have tested what they want in beta before propagating to prod?
Example: https://github.com/opensearch-project/.github/blob/main/.github/workflows/bake_time.yml#L17

.github/workflows/deploy-stacks.yml
rishabh6788: This specifically filters users with maintainer access.

.github/workflows/deploy-stacks.yml
rishabh6788: This or you can just let the workflow linger in the prod change set approval stage, i think the default timeout is 3-hr.

.github/workflows/deploy-stacks.yml
rishabh6788: I would rather have it break to make sure we always use the most recent LTS version.

rishabh6788: We are already lagging behind on cdk version.

.github/workflows/deploy-stacks.yml
rishabh6788: I'm okay if you want to remove it.

gaiksaya: Yes lets do that. Secrets are not visible once added in GH.

.github/workflows/deploy-stacks.yml
gaiksaya: Wasn't the case same here? https://github.com/opensearch-project/OpenSearch/blob/main/.github/workflows/benchmark-pull-request.yml#L118-L126
All admins also added for approval https://github.com/opensearch-project/OpenSearch/issues/17904 right?

.github/workflows/deploy-stacks.yml
gaiksaya: Right but that would be just good intentions. With bake time we would be enforcing the wait time since this would be affecting Prod system. We can add as an enhancement. Your call.

rishabh6788: I'm okay with this for now.

.github/workflows/deploy-stacks.yml
rishabh6788: fixed it by changing the `affiliation` parameter to `direct` instead of all. Can confirm it only lists maintainers for the repo.

gaiksaya (approved):

PR #572, opened 2025-06-10
move secrets logic to public stack

rishabh6788:
### Description
This PR moves the secrets creation logic to public ci-stack. The logic is such that it should not delete the existing secrets created using internal stack and only introduce new secrets.

### Issues Resolved

### DIFF
--- a/lib/secrets/secret-credentials.ts
+++ b/lib/secrets/secret-credentials.ts
@@ -1,0 +1,87 @@
+import { Stack, Tags } from 'aws-cdk-lib';
+import { ISecret, Secret } from 'aws-cdk-lib/aws-secretsmanager';
+import { OpenIdConnectProvider } from 'aws-cdk-lib/aws-iam';
+import { GitHubActionsFederateIntegrationForBranchesAndTags } from './gha-federate-access';
+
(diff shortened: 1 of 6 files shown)


lib/secrets/secret-credentials.ts
```diff
@@ -0,0 +1,87 @@
...
+    static snapshotsMavenUsername: ISecret;
+
+    static snapshotsMavenPassword: ISecret;
+
+    constructor(stack: Stack) {
+      AWSSecretsJenkinsCredentials.snapshotsMavenPassword = Secret.fromSecretNameV2(stack, 'imported-maven-snapshots-password', 'maven-snapshots-password');
+
+      AWSSecretsJenkinsCredentials.snapshotsMavenUsername = Secret.fromSecretNameV2(stack, 'imported-maven-snapshots-username', 'maven-snapshots-username');
+
+      AWSSecretsJenkinsCredentials.centralPortalMavenUsername = new Secret(stack, 'maven-central-portal-username', {
```
zelinh, line 20: Since we create new secrets this and below, will they be override every deployment? For above two secrets, seems like we are importing from existing.

peterzhuamazon (approved):

